PR #9, opened 2021-02-17
update docs/index.md for config file options. closes#8

bigdatasourav:


johnsmyth (approved):
I added and pushed some minor changes to the docs and the default config file.  Please review and make sure they look good before merging.  I also noticed that when no API token is set, we dont get an error from the plugin, the tables just return no results.  It might be better to explicitly return an error that says "'token'  must be set in the connection configuration" or something.

PR #47, opened 2021-09-06
Update the token format

francois2metz:
The github token format has been updated a few month ago. I updated the example format in the config file.


e-gineer:
Thanks @francois2metz ... nice catch!

We should also update the example token in `docs/index.md` I think to match?

francois2metz:
@e-gineer I missed that! Fixed.

misraved (approved):

e-gineer:
Thanks @francois2metz :-)

PR #165, opened 2022-06-05
Support teams, team repositories, and team members

japborst:
My first `go` and first steampipe contribution, so bear with me 😄 

**Changes**
Added 3 new tables:
- `github_team`
- `github_team_members`
- `github_team_repositories`

**Considerations**
I feel there's quite some duplicate code already present in the repo on how the list results are handled, but I feel that's outside the scope of this PR to fix.

**Open questions**
For some reason, the following works
```sql
SELECT
  t.id          AS team_id,
  t.name        AS team_name,
  t.privacy     AS team_privacy,
  t.description AS team_description,
  tr.name       AS repo_name,
  tr.fork       AS repo_is_fork,
  tr.private    AS repo_is_private,
  tr.archived   AS repo_is_archived,
  tr.language   AS repo_primary_language
FROM github.github_team AS t
  INNER JOIN github.github_team_repository AS tr
             ON t.organization = tr.organization
               AND t.slug = tr.slug
WHERE t.organization = 'my_org'
```

but this doesn't, even though it should work and I don't see clear difference between `github_team_repository` and `github_team_member`.
```sql

SELECT
  tm.login      AS user_login,
  t.id          AS team_id,
  t.name        AS team_name,
  t.privacy     AS team_privacy,
  t.description AS team_description
FROM github.github_team AS t
    INNER JOIN github.github_team_member AS tm
               ON t.organization = tm.organization
                 AND t.slug = tm.slug
WHERE t.organization = 'my_org'
>> [HV000] ERROR: rpc error: code = Internal desc = 'List' call for table 'github_team_member' is missing 1 required qual: column:'slug' operator: =
```


github/utils.go
```diff
@@ -92,7 +92,11 @@ func parseRepoFullName(fullName string) (string, string) {
 func convertTimestamp(ctx context.Context, input *transform.TransformData) (interface{}, error) {
 	switch t := input.Value.(type) {
 	case *github.Timestamp:
-		return t.Format(time.RFC3339), nil
+		if t == nil {
```
japborst, line 95: Open for better suggestions here. It can be `nil` so would be good to return `null` instead of fail the query

github/table_github_team_member.go
```diff
@@ -0,0 +1,146 @@
...
+		{Name: "role", Type: pb.ColumnType_STRING, Description: "The team member's role", Hydrate: tableGitHubTeamMemberGet},
+		{Name: "state", Type: pb.ColumnType_STRING, Description: "The membership state", Hydrate: tableGitHubTeamMemberGet},
+	}
+
+	return append(repoColumns, teamColumns...)
+}
+
+func tableGitHubTeamMember() *plugin.Table {
+	return &plugin.Table{
+		Name:        "github_team_member",
```
japborst, line 29: Terraform calls this `github_team_members` (plural), but given that everything in singular, it would make more sense to keep it that way IMHO.

github/table_github_team.go
```diff
@@ -0,0 +1,177 @@
...
+
+	client := connect(ctx, d)
+
+	type GetResponse struct {
+		team *github.Team
+		resp *github.Response
+	}
+
+	getDetails := func(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
+		detail, resp, err := client.Teams.GetTeamBySlug(ctx, org, slug)
```
japborst, line 161: Here and in the other tables I use the APIs based on `slug` as getting info by `id` is listed as `legacy` by the [API docs](https://docs.github.com/en/rest/teams/teams#get-a-team-legacy).

docs/tables/github_team.md
```diff
@@ -0,0 +1,32 @@
...
+The `github_team` table can be used to fetch team information for a given organization. **You must specify the organization** in the where or join clause (`where organization=`, `join github_team on organization=`).
+
+## Examples
+
+## Get all teams in an organization
+
+```sql
+select
+  t.id
+  t.name
```
misraved, line 12: This section of query needs to be updated to -
```sql
select
  id,
  name,
  privacy,
  description
from
  github.github_team
where 
  organization = 'my_org'

File: docs/tables/github_team_repository.md [hunk 5]
```diff
@@ -0,0 +1,50 @@
+# Table: github_team_repository
+
+A repository contains all of your project's files and each file's revision history.
+
+The `github_team_repository` table can be used to query information about repositories that a team have access to. **You must specify the organization and team slug** in the where or join clause (`where organization= AND slug=`, `join github_team_repository on organization= AND slug=`).
```
misraved, line 5:
```suggestion
The `github_team_repository` table can be used to query information about repositories that a team has access to. **You must specify the organization and team slug** in the where or join clause (`where organization= AND slug=`, `join github_team_repository on organization= AND slug=`).
```

github/table_github_team_member.go
misraved: Thanks @japborst for sticking to the singular naming convention as per our [docs](https://steampipe.io/docs/develop/standards#naming). `github_team_member` is correct 👍

github/table_github_team.go
misraved: @japborst you are absolutely correct to use the APIs based on `slug`, however since the majority of our queries and API calls are based on `id` (including the example queries), we might consider moving it as a part of breaking change.

japborst: @misraved perhaps we should then migrate all of them 🤔 How do you suggest moving forward?

github/utils.go
misraved: @japborst I believe that the default in the switch case takes care of the `nil` case.

However, can you think of any situation wherein the value of `t` would be `nil`?

github/table_github_team.go
misraved: @japborst The migration definitely makes sense since we should always look to update to the latest APIs.

Few points to be addressed:
- How many tables need to be updated to use the latest API format?
- We need to update the documents of the corresponding tables to include the latest changes.
-  The tables should still support querying by `id` and `slug` (tables should continue to require `organization` and at least one of `id` or `slug`, if both are passed in, prioritize `slug`). Updating the `github_my_team` table to use only `slug` seems like a hurried change since we would like to continue the support for `id` before we can phase it out completely.

Would be great to add those changes to this PR.

misraved:
Thanks @japborst for using Steampipe 👍. The tables and all the documents around it look great 😄 .

There are a few points which we would like to address before releasing them:
- Can we add `github_team_member` and `github_team_repository` tables information in the `github_team` table? By using joins with `github_user` and `github_repository` tables I believe we can retrieve similar information, instead of having two separate tables. This is similar to how we have implemented the `github_organization` and `github_my_organization` tables.
- As far as the open questions are concerned, I am also getting similar results. I am trying to dig in to figure out the possible reason for the failure. I will add more information once I have a conclusive reason for the failure.

Please let us know how you feel about the above approach, looking forward to your thoughts.

github/utils.go
japborst: I had to add this, so the case definitely exists 😄 

What happened was the following:
- The field was not present in the response
- The transform returns `nil`, passing to this function (from `transform.FromField("CreatedAt").Transform(convertTimestamp)`)

An alternative could be duplicate table definitions, but this seemed pretty sensible to support.

github/table_github_team.go
japborst: There are definitely more cases from the API docs, but looking at the code `*ById(...)` actually doesn't give us other results, so this is the only case 🎉 

I've pushed a commit that handles both cases and tested this works.

Supporting both is a little ugly, as all fields become optional to support both cases. Have a look at the get key columns. That also means less intuitive error messages.

Releasing without backwards compatibility would be preferred 😅

github/table_github_team.go
```diff
@@ -0,0 +1,205 @@
...
+		opt.Page = resp.NextPage
+	}
+
+	return nil, nil
+}
+
+//// HYDRATE FUNCTIONS
+
+func tableGitHubTeamGet(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
+	var org string
```
japborst, line 142: Here and below, the code to support both cases is pretty ugly. Suggestions are welcome.

japborst:
@misraved 
> Can we add github_team_member and github_team_repository tables information in the github_team table? 

I would prefer to not do that. Namely, if we collect to an array (repository names, member logins), that doesn't give us any additional information. I guess one _could_ flatten that array and join with the user table, but I'm not sure what we're really gaining then, as joining is definitely more convenient.

github/table_github_team.go
cbruno10: After discussing with @misraved and reviewing the code changes, I think it's probably ok to not make these changes backward compatible and to just support getting a team in either table by organization + slug. As long as we're not removing or changing the data for any columns, the changes originally proposed are ok.

cbruno10:
@japborst 
> I would prefer to not do that. Namely, if we collect to an array (repository names, member logins), that doesn't give us any additional information. I guess one _could_ flatten that array and join with the user table, but I'm not sure what we're really gaining then, as joining is definitely more convenient.

I took a closer look at the information available in the GitHub API around team memberships + team repository relationships, and it seems like there's some additional information available for each relationship that GitHub does not return from a regular list call that would be useful to include in a separate table representing the relationship.

For instance:
- For team members, there's `url`, `role`, and `state` available (https://docs.github.com/en/rest/teams/members#get-team-membership-for-a-user):
  ```
  {
  "url": "https://api.github.com/teams/1/memberships/octocat",
  "role": "maintainer",
  "state": "active"
  } 
  ```
- For team repository relationships, there's `permissions` available (https://docs.github.com/en/rest/teams/teams#check-team-permissions-for-a-repository):
  ```
  {
    "permissions": {
    "admin": false,
    "maintain": false,
    "push": false,
    "triage": false,
    "pull": true
  }
  ```

In some of the tables that @misraved had mentioned, like the `github_organization` and `github_repository` tables, we currently make separate hydrate calls to get organization members or repository collaborators (even though the Terraform provider represents them as separate resources), in an attempt to reduce the number of tables.

But, notably for organization members, we don't include each member's `url`, `role`, or `state` information anywhere in the table's columns, as we'd need to make an API call per member to the https://docs.github.com/en/rest/orgs/members#get-organization-membership-for-a-user API, which may be too expensive to do in the `github_organization` table.

So it seems like actually having separate tables for relationships can be beneficial, as there's information about the relationship that you need to make individual API calls for, e.g., one API call per organization member to get its state and role information. These types of calls don't seem great to include in the parent table, like `github_team` or `github_organization`.

We could follow what Terraform does in general for these types of tables, which would mean the two additional tables in this PR would be named correctly. However, @japborst, if we were to proceed in this way, one suggestion I'd have is to reduce the columns down in both. For instance, in the `github_team_member` table, I'd propose that the columns should only include fields from the responses in https://docs.github.com/en/rest/teams/members#list-team-members and https://docs.github.com/en/rest/teams/members#get-team-membership-for-a-user, but not necessarily the full set of columns from the `github_team` and `github_user` tables, as there are additional columns in those tables that we don't get from the API responses.

@johnsmyth @japborst Interested in your thoughts on the proposal above (it came out a lot longer than I thought it'd be). Thanks!

github/table_github_team.go
japborst: Clear 👍  The exposed data indeed doesn't change. The team `id` and `organization_id` are still there, for people relying on it.

I've reverted my last commit

github/utils.go
japborst: Just doing `NullIfZero()` already did the trick, thanks 🙏

japborst:
Thanks, @cbruno10! Agree to your points. Especially `role` is quite relevant.
- I do see that the repo permissions are not here yet, where actually that's super useful information. Will see if I can add easily.
- I've updated the `github_team_member` column definition to narrow the scope, as you mentioned.
- The `github_my_team` and `github_team` data is the same, as it uses the same underlying API, so no changes in columns there
- Same goes for the `github_team_repository` and `github_repository` columns

github/table_github_user.go
```diff
@@ -10,6 +10,42 @@ import (
 	"github.com/turbot/steampipe-plugin-sdk/v3/plugin/transform"
 )
 
+func gitHubUserColumns() []*plugin.Column {
```
cbruno10, line 13: Do we still need this change if we're not using these columns in the `github_team_member` table anymore?

github/table_github_team_member.go
```diff
@@ -0,0 +1,163 @@
...
+		{Name: "organizations_url", Type: pb.ColumnType_STRING, Description: "The URL of the user's organizations."},
+		{Name: "received_events_url", Type: pb.ColumnType_STRING, Description: "The URL of the user's received events."},
+		{Name: "repos_url", Type: pb.ColumnType_STRING, Description: "The URL of the user's repos."},
+		{Name: "site_admin", Type: pb.ColumnType_BOOL, Description: "If true, user is an administrator."},
+		{Name: "starred_url", Type: pb.ColumnType_STRING, Description: "The URL of the user's stars."},
+		{Name: "subscriptions_url", Type: pb.ColumnType_STRING, Description: "The URL of the user's subscriptions."},
+		{Name: "type", Type: pb.ColumnType_STRING, Description: "The type of account."},
+		{Name: "url", Type: pb.ColumnType_STRING, Description: "The URL of the user."},
+
+		{Name: "role", Type: pb.ColumnType_STRING, Description: "The team member's role", Hydrate: tableGitHubTeamMemberGet},
```
cbruno10, line 39:
```suggestion
		{Name: "role", Type: pb.ColumnType_STRING, Description: "The team member's role.", Hydrate: tableGitHubTeamMemberGet},
```

Please ensure that all table and column descriptions end with periods (as per https://steampipe.io/docs/develop/standards#table-and-column-descriptions).

github/table_github_team_member.go
```diff
@@ -0,0 +1,163 @@
...
+		{Name: "state", Type: pb.ColumnType_STRING, Description: "The membership state", Hydrate: tableGitHubTeamMemberGet},
```
cbruno10, line 40:
```suggestion
		{Name: "state", Type: pb.ColumnType_STRING, Description: "The membership state.", Hydrate: tableGitHubTeamMemberGet},
```

github/table_github_team_repository.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	return append(repoColumns, teamColumns...)
+}
+
+func tableGitHubTeamRepository() *plugin.Table {
+	return &plugin.Table{
+		Name:        "github_team_repository",
+		Description: "GitHub Repositories that a given team is associated with. GitHub Repositories contain all of your project's files and each file's revision history.",
+		List: &plugin.ListConfig{
```
cbruno10, line 29: Is it possible to add a Get function here, which takes in an organization, slug, and repository full name? You may be able to use this API - https://docs.github.com/en/rest/teams/teams#check-team-permissions-for-a-repository, though in its sample response, it looks like it has more fields than its list counterpart (https://docs.github.com/en/rest/teams/teams#list-team-repositories).

github/table_github_team_repository.go
```diff
@@ -0,0 +1,101 @@
...
+			KeyColumns: []*plugin.KeyColumn{
+				{Name: "organization", Require: plugin.Required},
+				{Name: "slug", Require: plugin.Required},
+			},
+			Hydrate:           tableGitHubTeamRepositoryList,
+			ShouldIgnoreError: isNotFoundError([]string{"404"}),
+		},
+		Columns: gitHubTeamRepositoryColumns(),
```
cbruno10, line 37: As discussed, please add a `permissions` as well if possible.

cbruno10 (changes requested):
@japborst I'm also looking into why your second query is failing you mentioned in your original pull request comment, and what's a solution/workaround to allow those queries to work properly. I'll add another comment once I have more findings.

In the meantime, I've added some more suggestions, and if you have any questions please let me know, thanks!

github/table_github_user.go
japborst: It makes it more consistent, but technically not indeed. I'll revert it 👍

japborst:
@cbruno10 @misraved I've addressed all the comments.

cbruno10:
Hey @japborst , I was looking at how we could get your second example working (joining `github_team` and `github_team_member`), as there's a current limitation from Steampipe's side around query planning joins on tables with multiple required key quals (which `github_team` and `github_team_member` both have).

I believe that adding a `ParentHydrate` to `github_team`, with a user's organizations acting as the parent hydrate, should help solve this issue and make the table a bit easier to use.

So now if a user runs `select * from github_team`, the following sequence occurs:
- `tableGitHubMyOrganizationList` is called first, listing all organizations the user is a member of
- For each organization, `tableGitHubTeamList` is called, listing all teams in the organization the user has visibility to
- If any columns are included in the query that rely on the hydrate `tableGitHubTeamGet`, `tableGitHubTeamGet` will be called for each team

If a user runs `select * from github_team where organization = 'myorg' and slug = 'team-slug'`, only the `tableGitHubTeamGet` hydrate function will be called (same as before).

The `github_my_team` table is unaffected by these proposed changes.

Because users can only list teams they're either a part of or have visibility to (which I believe requires the user to at least be a member of an organization they're attempting to list teams for), having the `github_team` table first list all organizations that user is a member of should give an accurate list of teams they have visibility to across GitHub.

I've implemented and tested these changes in https://github.com/turbot/steampipe-plugin-github/tree/add-parent-hydrate-team, and they appear to be working well so far (your original query now runs ok as well). Another change I made as part of that branch was to bump go-github to v45, like you had done in the audit logs PR, as I needed another response property for the `Team` item.

Can you please have a look at my proposal above along with the branch, and let me know if this is aligned with your original intentions for this PR, and if you have any other feedback/suggestions? Thanks!

japborst:
Nice @cbruno10! That works. I've pushed some final changes.

Good to go from my end ✅

cbruno10:
Thanks so much @japborst for all of the hard work and great discussions!

PR #166, opened 2022-06-11
Support GitHub audit logs

### LINKED ISSUE #164: Query Github Audit Logs - Steampipe
**Is your feature request related to a problem? Please describe.**

Create a new table for Github Audit logs for Steampipe. I use this query in our internal tooling to find policy overrides like ["Branch Protection Overrides"](https://docs.github.com/en/organizations/keeping-your-organization-secure/managing-security-settings-for-your-organization/reviewing-the-audit-log-for-your-organization#protected_branch-category-actions) to trigger a notification. Would be nice to have this data within Steampipe for us to start using it for security notifications. In addition the Github Audit Logs seems to have other important data like "tokens created, OAuth app creations" which would be useful for finding security gaps. 

curl \
  -H "Accept: application/vnd.github.v3+json" \
  -H "Authorization: token ghp_...(redacted)" \
  --request GET \
  "https://api.github.com/orgs/loomhq/audit-log?phrase=created:2022-05-10&page=1&per_page=100"

**Describe the solution you'd like**

Able to query github audit logs (https://github.com/organizations/loomhq/settings/audit-log) for the last X days. Differentiating records based on github usernames, event types and timestamps so that it is easy to query and filter results.

japborst:
Fixes #164 

Unfortunately not able to test myself, as this requires Enterprise access. However, based on the export from the UI I've verified the fields.

To fetch audit logs using the API, required an upgrade to the `go-github` package. I can also do that in a separate PR, if that's preferred.

### DIFF
--- a/docs/tables/github_audit_log.md
+++ b/docs/tables/github_audit_log.md
@@ -1,6 +1,6 @@
 # Table: github_audit_log
 
-The `github_audit_log` table helps to find all audit events for an organization.
+The `github_audit_log` table helps to find all audit events for an organization. Note: this only works for organizations on an **GitHub Enterprise plan**.
 
 **You must always specify the organization** in the where or join clause using the `organization` column. Additionally, you can filter the logs by using a search phrase (`phrase`), event types (`include`), and before/after a timestamp (`created_at`).
 
@@ -19,6 +19,7 @@ from
   github_audit_log
 where
   organization = 'my_org'
+order by created_at
 limit 10;
 ```
 
--- a/docs/tables/github_audit_log.md
+++ b/docs/tables/github_audit_log.md
@@ -6,22 +6,7 @@ The `github_audit_log` table helps to find all audit events for an organization.
 
 ## Examples
 
-### Get all audit events for an organization
-
-```sql
-select
-  id,
-  created_at,
-  actor,
-  action,
-  data
-from
-  github_audit_log
-where
-  organization = 'my_org';
-```
-
-### Get specific audit events
+### Get recent audit events for an organization
 
 ```sql
 select
@@ -34,6 +19,42 @@ from
   github_audit_log
 where
   organization = 'my_org'
-  and phrase = "action:repo.create action:repo.destroy"
-  and created_at >= '2022-01-01';
+limit 10;
+```
+
+### Get specific audit events
+
+For example, find out which repos have been created or deleted on the first of January.
+
+```sql
+select
+  id,
+  created_at,
+  actor,
+  action,
+  data
+from
+  github_audit_log
+where
+  organization = 'my_org'
+  and action IN ('repo.create', 'repo.destroy')
+  and created_at = '2022-01-01';
+```
+
+### Get specific events by a specific actor (user) in the last 30 days
+
+```sql
+select
+  id,
+  created_at,
+  actor,
+  action,
+  data
+from
+  github_audit_log
+where
+  organization = 'my_org'
+  and actor = 'some_user'
+  and created_at > (created_at - interval '30' day)
+order by created_at;
 ```
--- a/docs/tables/github_audit_log.md
+++ b/docs/tables/github_audit_log.md
@@ -16,9 +16,9 @@ select
   action,
   data
 from
-    github_audit_log
+  github_audit_log
 where
-    organization = 'my_org'
+  organization = 'my_org';
 ```
 
 ### Get specific audit events
@@ -31,9 +31,9 @@ select
   action,
   data
 from
-    github_audit_log
+  github_audit_log
 where
-    organization = 'my_org'
-    and phrase = "action:repo.create action:repo.destroy"
-    and created_at >= '2022-01-01'
+  organization = 'my_org'
+  and phrase = "action:repo.create action:repo.destroy"
+  and created_at >= '2022-01-01';
 ```
--- a/docs/tables/github_audit_log.md
+++ b/docs/tables/github_audit_log.md
@@ -16,7 +16,7 @@ select
   action,
   data
 from
-    github.github_audit_log
+    github_audit_log
 where
     organization = 'my_org'
 ```
@@ -31,7 +31,7 @@ select
   action,
   data
 from
-    github.github_audit_log
+    github_audit_log
 where
     organization = 'my_org'
     and phrase = "action:repo.create action:repo.destroy"
--- a/docs/tables/github_audit_log.md
+++ b/docs/tables/github_audit_log.md
@@ -1,0 +1,39 @@
+# Table: github_audit_log
+
+The `github_audit_log` table helps to find all audit events for an organization.
+
+**You must always specify the organization** in the where or join clause using the `organization` column. Additionally, you can filter the logs by using a search phrase (`phrase`), event types (`include`), and before/after a timestamp (`created_at`).
+
+## Examples
+
+### Get all audit events for an organization
+
+```sql
+select
+  id,
+  created_at,
+  actor,
+  action,
+  data
+from
+    github.github_audit_log
+where
+    organization = 'my_org'
+```
+
+### Get specific audit events
+
+```sql
+select
+  id,
+  created_at,
+  actor,
+  action,
+  data
+from
+    github.github_audit_log
+where
+    organization = 'my_org'
+    and phrase = "action:repo.create action:repo.destroy"
+    and created_at >= '2022-01-01'
+```
--- a/github/table_github_audit_log.go
+++ b/github/table_github_audit_log.go
@@ -60,6 +60,8 @@ func tableGitHubAuditLogList(ctx context.Context, d *plugin.QueryData, h *plugin
 		ListCursorOptions: github.ListCursorOptions{PerPage: 100},
 	}
 
+	// TODO: Support quals["action"] to filter using the phrase parameter
+
 	if quals["created_at"] != nil {
 		for _, q := range d.Quals["created_at"].Quals {
 			givenTime := q.Value.GetTimestampValue().AsTime().Format(time.RFC3339)
@@ -69,6 +71,8 @@ func tableGitHubAuditLogList(ctx context.Context, d *plugin.QueryData, h *plugin
 				op = ""
 			}
 
+			// TODO: Handle BETWEEN clause
+			// TODO: Handle from/to to use created:<fromt>...<to> instead
 			phrase += " created:" + op + givenTime
 			opts.Phrase = &phrase
 		}
--- a/github/table_github_audit_log.go
+++ b/github/table_github_audit_log.go
@@ -21,6 +21,7 @@ func tableGitHubAuditLog(ctx context.Context) *plugin.Table {
 				{Name: "organization", Require: plugin.Required},
 				{Name: "phrase", Require: plugin.Optional},
 				{Name: "include", Require: plugin.Optional},
+				{Name: "action", Require: plugin.Optional},
 				{Name: "created_at", Require: plugin.Optional, Operators: []string{">", ">=", "<", "<=", "="}},
 			},
 			Hydrate: tableGitHubAuditLogList,
@@ -60,8 +61,6 @@ func tableGitHubAuditLogList(ctx context.Context, d *plugin.QueryData, h *plugin
 		ListCursorOptions: github.ListCursorOptions{PerPage: 100},
 	}
 
-	// TODO: Support quals["action"] to filter using the phrase parameter
-
 	if d.Quals["created_at"] != nil {
 		for _, q := range d.Quals["created_at"].Quals {
 			givenTime := q.Value.GetTimestampValue().AsTime().Format(time.RFC3339)
@@ -74,6 +73,11 @@ func tableGitHubAuditLogList(ctx context.Context, d *plugin.QueryData, h *plugin
 			phrase += " created:" + op + givenTime
 			opts.Phrase = &phrase
 		}
+	}
+
+	if quals["action"] != nil {
+		phrase += " action:" + quals["action"].GetStringValue()
+		opts.Phrase = &phrase
 	}
 
 	type ListPageResponse struct {
--- a/github/table_github_audit_log.go
+++ b/github/table_github_audit_log.go
@@ -84,8 +84,8 @@ func tableGitHubAuditLogList(ctx context.Context, d *plugin.QueryData, h *plugin
 	}
 
 	type ListPageResponse struct {
-		result []*github.AuditEntry
-		resp   *github.Response
+		entries []*github.AuditEntry
+		resp    *github.Response
 	}
 
 	client := connect(ctx, d)
@@ -99,15 +99,15 @@ func tableGitHubAuditLogList(ctx context.Context, d *plugin.QueryData, h *plugin
 	}
 
 	listPage := func(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
-		result, resp, err := client.Organizations.GetAuditLog(ctx, org, opts)
+		entries, resp, err := client.Organizations.GetAuditLog(ctx, org, opts)
 
 		if err != nil {
 			return nil, err
 		}
 
 		return ListPageResponse{
-			result: result,
-			resp:   resp,
+			entries: entries,
+			resp:    resp,
 		}, nil
 	}
 
@@ -119,7 +119,7 @@ func tableGitHubAuditLogList(ctx context.Context, d *plugin.QueryData, h *plugin
 		}
 
 		listResponse := listPageResponse.(ListPageResponse)
-		auditResults := listResponse.result
+		auditResults := listResponse.entries
 		resp := listResponse.resp
 
 		for _, i := range auditResults {
@@ -131,11 +131,11 @@ func tableGitHubAuditLogList(ctx context.Context, d *plugin.QueryData, h *plugin
 			}
 		}
 
-		if resp.NextPage == 0 {
+		if resp.After == "" {
 			break
 		}
 
-		opts.ListCursorOptions.Page = resp.NextPageToken
+		opts.After = resp.After
 	}
 
 	return nil, nil
--- a/github/table_github_audit_log.go
+++ b/github/table_github_audit_log.go
@@ -31,11 +31,11 @@ func tableGitHubAuditLog(ctx context.Context) *plugin.Table {
 			{Name: "include", Type: pb.ColumnType_STRING, Transform: transform.FromQual("include"), Description: "The event types to include: web, git, all."},
 
 			// Top columns.
-			{Name: "id", Type: pb.ColumnType_STRING, Description: "The id of the audit event.", Transform: transform.FromField("_document_id")},
-			{Name: "created_at", Type: pb.ColumnType_TIMESTAMP, Description: "The timestamp of the audit event.", Transform: transform.FromField("created_at").Transform(convertTimestamp)},
+			{Name: "id", Type: pb.ColumnType_STRING, Description: "The id of the audit event.", Transform: transform.FromField("DocumentID")},
+			{Name: "created_at", Type: pb.ColumnType_TIMESTAMP, Description: "The timestamp of the audit event.", Transform: transform.FromField("CreatedAt").Transform(convertTimestamp)},
 			{Name: "action", Type: pb.ColumnType_STRING, Description: "The action performed."},
 			{Name: "actor", Type: pb.ColumnType_STRING, Description: "The GitHub user who performed the action."},
-			{Name: "actor_location_country_code", Type: pb.ColumnType_STRING, Description: "The country location of the actor at the moment of the action.", Transform: transform.FromField("actor_location.country_code")},
+			{Name: "actor_location_country_code", Type: pb.ColumnType_STRING, Description: "The country location of the actor at the moment of the action.", Transform: transform.FromField("ActorLocation.CountryCode")},
 
 			// Optional columns, depending on the audit event.
 			{Name: "team", Type: pb.ColumnType_STRING, Description: "The GitHub team, when the action relates to a team."},
--- a/github/table_github_audit_log.go
+++ b/github/table_github_audit_log.go
@@ -62,7 +62,7 @@ func tableGitHubAuditLogList(ctx context.Context, d *plugin.QueryData, h *plugin
 
 	// TODO: Support quals["action"] to filter using the phrase parameter
 
-	if quals["created_at"] != nil {
+	if d.Quals["created_at"] != nil {
 		for _, q := range d.Quals["created_at"].Quals {
 			givenTime := q.Value.GetTimestampValue().AsTime().Format(time.RFC3339)
 
@@ -71,8 +71,6 @@ func tableGitHubAuditLogList(ctx context.Context, d *plugin.QueryData, h *plugin
 				op = ""
 			}
 
-			// TODO: Handle BETWEEN clause
-			// TODO: Handle from/to to use created:<fromt>...<to> instead
 			phrase += " created:" + op + givenTime
 			opts.Phrase = &phrase
 		}
@@ -94,6 +92,7 @@ func tableGitHubAuditLogList(ctx context.Context, d *plugin.QueryData, h *plugin
 	}
 
 	listPage := func(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
+		plugin.Logger(ctx).Debug("tableGitHubAuditLogs", "org", org, "include", include, "phrase", *opts.Phrase)
 		entries, resp, err := client.Organizations.GetAuditLog(ctx, org, opts)
 
 		if err != nil {
--- a/github/table_github_audit_log.go
+++ b/github/table_github_audit_log.go
@@ -5,7 +5,7 @@ import (
 	"time"
 
 	"github.com/google/go-github/v45/github"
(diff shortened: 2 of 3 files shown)


misraved:
Thanks @japborst for adding the `github_audit_log` table 👍 .

I took the first pass at it and the changes look promising 😄 , however, we don't have enterprise access to test it out. I am hoping @ngowraj will be able to help us out here.

As far as the upgrade to the latest version of the `go-github` package is concerned, the changes look good. Even though the CHANGELOG of the `go-github` package doesn't suggest any breaking change, @japborst did you encounter any error while testing out the upgrade/changes across other GitHub tables?

japborst:
I checked _some_ tables, but not _all_. There's no automated testing, so that doesn't help either 😅 

I can try running all the example queries from the docs, to see if those don't return an error.

japborst:
@misraved I've ran all samples, and they run just fine ✅ 

Effectively ran everything returned using 
```console
$ cat docs/tables/*.md | sed -n '/```/,/```/p' | sed 's/```sql//' | sed 's/```//'
```

ngowraj:
@japborst I am new to Steampipe. I would appreciate your help in testing out this change. 

For testing out the table `github_audit_log`, I believe I need to follow the instructions [here](https://github.com/turbot/steampipe-plugin-github#developing). `Make` command has to be done on your remote branch `japborst:japborst/github-audit-log ` and not the `main` branch since the code is not merged into main. 

I tried cloning the repo. Doing a `git fetch` and then `git checkout japborst/github-audit-log ` but this fails with error
```
error: pathspec 'japborst/github-audit-log' did not match any file(s) known to git
```

I was wondering if I am missing something? Appreciate your help here. Thanks!

cc @misraved

misraved:
Thanks @ngowraj for testing out the changes.  Looks like you have already cloned the github plugin repository, you can now refer the following steps to fetch the branch of @japborst :

- Run the git commands to switch to the forked branch:
```
git remote add fork1 https://github.com/japborst/steampipe-plugin-github.git
git fetch fork1
git checkout fork1/japborst/github-audit-log
```
- Once you have switched to the branch, you can then run the `make` command to build the changes locally.
- Run the `steampipe query` command to open the interactive query session.
- Run `.inspect github_audit_log` to get a view of the columns available in the `github_audit_log` table which should look something like this - 
```
inspect github_audit_log
+-----------------------------+--------------------------+-----------------------------------------------------------------+
| column                      | type                     | description                                                     |
+-----------------------------+--------------------------+-----------------------------------------------------------------+
| _ctx                        | jsonb                    | Steampipe context in JSON form, e.g. connection_name.           |
| action                      | text                     | The action performed.                                           |
| actor                       | text                     | The GitHub user who performed the action.                       |
| actor_location_country_code | text                     | The country location of the actor at the moment of the action.  |
| created_at                  | timestamp with time zone | The timestamp of the audit event.                               |
| data                        | jsonb                    | Additional data relating to the audit event.                    |
| id                          | text                     | The id of the audit event.                                      |
| include                     | text                     | The event types to include: web, git, all.                      |
| organization                | text                     |                                                                 |
| phrase                      | text                     | The search phrase for your audit events.                        |
| repo                        | text                     | The GitHub repository, when the action relates to a repository. |
| team                        | text                     | The GitHub team, when the action relates to a team.             |
| user                        | text                     | The GitHub user, when the action relates to a user.             |
+-----------------------------+--------------------------+-----------------------------------------------------------------+
```

Once again thanks @ngowraj for attempting to test the changes 👍 . Please let me know if you need some help with the above steps 👍 .

cbruno10:
@ngowraj Alternatively, you can clone `japborst:japborst/github-audit-log` in a new directory and run all commands from there to avoid the need of creating a new remote:
```
cbruno@M1P steampipe % mkdir japborst
cbruno@M1P steampipe % cd japborst
cbruno@M1P japborst % git clone -b japborst/github-audit-log git@github.com:japborst/steampipe-plugin-github.git .
Cloning into '.'...
remote: Enumerating objects: 1338, done.
remote: Counting objects: 100% (567/567), done.
remote: Compressing objects: 100% (197/197), done.
remote: Total 1338 (delta 452), reused 436 (delta 367), pack-reused 771
Receiving objects: 100% (1338/1338), 473.67 KiB | 5.57 MiB/s, done.
Resolving deltas: 100% (894/894), done.
cbruno@M1P japborst % make
go build -o ~/.steampipe/plugins/hub.steampipe.io/plugins/turbot/github@latest/steampipe-plugin-github.plugin *.go
go: downloading github.com/google/go-github/v45 v45.1.0
cbruno@M1P japborst % steampipe query
Welcome to Steampipe v0.14.6
For more information, type .help
> ...
```

docs/tables/github_audit_log.md
```diff
@@ -0,0 +1,39 @@
...
+
+```sql
+select
+  id,
+  created_at,
+  actor,
+  action,
+  data
+from
+    github_audit_log
```
misraved, line 19:
```suggestion
   github_audit_log
```

docs/tables/github_audit_log.md
```diff
@@ -0,0 +1,39 @@
...
```
misraved, line 34:
```suggestion
   github_audit_log
```

docs/tables/github_audit_log.md
```diff
@@ -0,0 +1,39 @@
...
+where
+    organization = 'my_org'
+    and phrase = "action:repo.create action:repo.destroy"
+    and created_at >= '2022-01-01'
```
misraved, line 38:
```suggestion
  and created_at >= '2022-01-01'
```

cbruno10:
@japborst We have a utility function which may be helpful for converting the timestamps - https://github.com/turbot/steampipe-plugin-github/blob/cfd72280fa447e1662a65b10728f3655bc483a61/github/utils.go#L92-L101

Here are two places (among others) that we use it to transform a column - https://github.com/turbot/steampipe-plugin-github/blob/63da70c695f996b81adce89f085745ac1371b169/github/table_github_user.go#L45-L46

japborst:
Ah, sorry about that @ngowraj!

I've pushed a commit that should ifx that. Can you pull the latest changes and try again?
```console
$ git checkout japborst/github-audit-log
$ git pull
$ make
```

misraved:
@japborst could you please work on removing the merge conflicts? I believe it will unblock @ngowraj for testing out the ne table. Thanks 👍

japborst:
@misraved sure. Done ✅

ngowraj:
I will test this out by this week. Thanks!

ngowraj:
It works. I was able to query for the first few recent audit logs using SQL. 
`select * from github_audit_log where organization = 'loomhq' limit 10;`

I would like to get your thoughts on the following and see if we can get the following as part of this integration. 

1. Field like `phrase, include, id, created_at, country_code, team, user, data` are mostly or always null. 
Screen Shot 2022-06-22 at 4 18 26 PM

2. Discrepancy in data returned. A REST call to api.github.com to fetch all logs for today (June 22, 2022) for actor ngowraj (myself) returns more than 25+ records with non-null values for `created_at` field.
```
curl -H "Accept: application/vnd.github.v3+json" -H "Authorization: token ghp_<redacted> --request GET "https://api.github.com/orgs/loomhq/audit-log?phrase=actor:ngowraj+created:2022-06-22&page=1&per_page=100"
``` 

Whereas when I try to get the same data from Steampipe I get only 2 records (screenshot above) and the `created_at` value is null. 

Query is `select * from github_audit_log where organization = 'loomhq' and actor = 'ngowraj';`

I believe this happens since we only have the first 100 records in our table at any point of time. `select count(*) from github_audit_log where organization = 'loomhq';` returns **100** always. Is there a way for us paginate and get more records into the `github_audit_log` table or make a REST call with filters when we have multiple conditions within our `where` clause. Not sure what would be the best strategy here. I believe the team has more context on how we can get this done.

This is kind of important since events like [pull_request_review](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_review) are very common and mostly innocuous and events like `org_credential_authorization.grant` which needs care and attention are pushed beyond the first 100 records very frequently.

japborst:
Thanks @ngowraj for testing! 💪 

That would indicate that pagination is indeed not working as it should! A bit odd, as pagination was implemented. Will see if I can find something.

As FYI
```sql
select * from github_audit_log where organization = 'loomhq' and actor = 'ngowraj';
```

First queries everything `where organization = 'loomhq' and then after loading the results filters for the actor. If we only load 100 results in total, this explains why you only get 2 results when filtering for the actor.

`created_at` shouldn't be `null` indeed, and might also explain why you're not getting any results when filtering for it.

ngowraj:
@japborst Makes sense. 

> First queries everything `where organization = 'loomhq' and then after loading the results filters for the actor. If we only load 100 results in total, this explains why you only get 2 results when filtering for the actor.

Makes sense. Do you think we can remove `organization` column to be mandatory? That would be helpful since all the audit logs for `loomhq` have the same value `loomhq`. If this possible then I can load records for actor = 'ngowraj' to get the first 100 records and then filter based on actions or created_at. 

```
select * from github_audit_log where organization = 'loomhq' and actor = 'ngowraj';
```

> created_at shouldn't be null indeed, and might also explain why you're not getting any results when filtering for it.
👍 

>  A bit odd, as pagination was implemented. Will see if I can find something.
Thanks 🙏

cbruno10:
> Makes sense. Do you think we can remove `organization` column to be mandatory? That would be helpful since all the audit logs for `loomhq` have the same value `loomhq`. If this possible then I can load records for actor = 'ngowraj' to get the first 100 records and then filter based on actions or created_at.

@ngowraj I think keeping `organization` as a required key column is worthwhile for the audit log table in particular. For some organization level resources, like `github_team`, we've actually used a [parent hydrate](https://github.com/turbot/steampipe-plugin-github/blob/main/github/table_github_team.go#L51-L55), so the table's list will first list all organizations you're a part of, and then all of the teams for each organization.

But for audit logs, especially considering how many entries there could be per organization, it feels like keeping `organization` as a required key column would help users avoid hitting GitHub's rate limits accidentally, e.g., a user running `select * from github_audit_log` as a first test could push them closer to the hourly rate limit.

Beside adding `where organization = 'loomhq'` to each query, are there any other factors impacting you?

@japborst Thanks again for working so much on this table! Please let us know if there's anything we can do to help as well.

ngowraj:
> Beside adding where organization = 'loomhq' to each query, are there any other factors impacting you?

Not really. I am fine with adding organization = 'loomhq' to every query. 

With this capability of querying audit logs, I am trying to have the following query primarily to return the right results. Currently, I am not able to do it but once pagination is fixed by @japborst I think it should be doable. 

Example => Get actions like `org_credential_authorization.grant`, `protected_branch.policy_override` filter by created_at date and actor. 

Example query would be
```
select count(*) as cnt, organization, action, actor, created_at from github_audit_log where organization = 'loomhq' and action in ('org_credential_authorization.grant', 'protected_branch.policy_override') and DATE(created_at) > (NOW() - INTERVAL 1 DAY) group by organization, action, actor, created_at order by cnt desc;
```

japborst:
@ngowraj

> Field like phrase, include, id, created_at, country_code, team, user, data are mostly or always null.
- `phrase` and `include` are optional query parameters you can provide.
-  `id`, `created_at` and `actor_location_country_code` shouldn't be `null`, **that should be fixed now**.
- `team`, `user`, and `data` are optional fields. `team` and `user` are only returned when the action is about a user or team (e.g. team creation, or user invite to the org)


Next to the above, I've pushed a few fixes:
1. Pagination should work now 🤞 
2. `created_at` filtering was not done properly; that should be fixed.

>Example query would be
>```sql
>select count(*) as cnt, organization, action, actor, created_at from github_audit_log where organization = 'loomhq' and action in ('org_credential_authorization.grant', 'protected_branch.policy_override') and DATE(created_at) > (NOW() - INTERVAL 1 DAY) group by organization, action, actor, created_at order by cnt desc;
>```

Note that it would first fetch all logs for the org and date, and only once retrieved filter the actions. These are currently not passed in as a search parameter. Something that can be improved later.

ngowraj:
I will test it out today.

ngowraj:
Works great. Appreciate your effort and time 🙏

ngowraj:
@japborst @misraved 

**What am I trying?**
I am trying to have an incremental query which should return results **_fast_**. Example => Query for logs with certain filters for an organization within a start and end period. 

**Queries**
Basic query` select * from github_audit_log where organization = 'loomhq' limit 5;` works great. 

1. Now, trying incremental query to fetch results for the last 2 days. 
```
select * from github_audit_log where organization = 'loomhq' and DATE(created_at) > (NOW() - INTERVAL 2 DAY) limit 10;
```
**Error: syntax error at or near "2" (SQLSTATE 42601)**

2. A different query  to fetch results for the last 2 days. 
```
select * from github_audit_log where organization = 'loomhq' and created_at between '2022-06-27' and '2022-06-29' limit 10;
```
**Empty results. No records returned. We do have a plenty of logs between the dates mentioned above.**

If I fire the same query with an additional filter `and action = 'protected_branch.policy_override' ` then it does return the results. So not sure why I am not able to filter out on a particular period alone

3. Interestingly created_at greater (>) works but lesser (<) doesn't work. 
```
select * from github_audit_log where organization = 'loomhq' and created_at < '2022-06-28' limit 10; 
```

**No results. Whereas created_at > '2022-06-28' returns the most recent ones.** 

**My Goal**
Have a fast query returning records filtered based on action and group by actor. Something like this.

`select count(*) as cnt, actor from github_audit_log where organization = 'loomhq' and created_at between '2022-06-28' and '2022-06-29' and action = 'protected_branch.policy_override' group by actor;`

This query says "Loading Results" for more than 30 mins and I had to abort it. 

Any thoughts on how I can fetch the results that I am looking for?

japborst:
Thanks for the clear overview @ngowraj!

A few things before diving into the topics:
1. Can you confirm pagination does work now? i.e. there are more than 100 results
2. Can you confirm these fields are not `null` any more? `id`, `created_at` and `actor_location_country_code`

wrt the date filters:
- `DATE(created_at) > (NOW() - INTERVAL 2 DAY)` AFAIK that's not supported by steampipe, but perhaps @misraved knows how to support that
- `created_at between '2022-06-27' and '2022-06-29'` this isn't supported in other similar plugins, but this should be possible I think
- created_at greater (>) works but lesser (<) doesn't work: that's not expected. I may have an idea, but can't test so not ideal 😅 

The reason why the action filter is slow, is because that is only done _after_ fetching _all_ results. Will take a look if we can support that as search query (similar to the UI).

cbruno10:
@ngowraj For this query:
```sql
select
  * 
from
  github_audit_log 
where
  organization = 'loomhq' 
  and DATE(created_at) > (NOW() - INTERVAL 2 DAY) limit 10;
```

Can you please try:
```sql
select
  * 
from
  github_audit_log 
where
  organization = 'loomhq' 
  and created_at > (created_at - interval '2' day)
limit 10;
```

@japborst For:
```sql
select
  * 
from
  github_audit_log 
where
  organization = 'loomhq' 
  and created_at between '2022-06-27' and '2022-06-29'
limit 10;
```

I believe Steampipe should handle the `between` operator correctly, but I'm not 100% sure on that. In the database or plugin logs, I believe there are some messages (maybe at DEBUG or TRACE level) that report what quals Steampipe sees, including the operators. I'll also check from my end to see if I can test this in another plugin table that supports date column key quals.

japborst:
Thanks @cbruno10. I've added some tracing, and actually `between` is interpreted as `>=` and `<=` so that's perfect.

Found a bug, however, which causes the issue. Will push a fix soon.

ngowraj:
Will test it out.

japborst:
Also added support for filtering on `actor` through the API.

japborst:
In the mean time, would appreciate a code review @misraved or @cbruno10 :)

github/table_github_audit_log.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/turbot/steampipe-plugin-sdk/v3/plugin"
+	"github.com/turbot/steampipe-plugin-sdk/v3/plugin/transform"
+)
+
+//// TABLE DEFINITION
+
+func tableGitHubAuditLog(ctx context.Context) *plugin.Table {
+	return &plugin.Table{
+		Name:        "github_audit_log",
+		Description: "Gets the audit log for an organization.",
```
cbruno10, line 18:
```suggestion
		Description: "Gets the audit logs for an organization.",
```

Minor change, but plural reads a bit better.

github/table_github_audit_log.go
```diff
@@ -0,0 +1,146 @@
...
+				{Name: "phrase", Require: plugin.Optional},
+				{Name: "include", Require: plugin.Optional},
+				{Name: "action", Require: plugin.Optional},
+				{Name: "actor", Require: plugin.Optional},
+				{Name: "created_at", Require: plugin.Optional, Operators: []string{">", ">=", "<", "<=", "="}},
+			},
+			Hydrate: tableGitHubAuditLogList,
+		},
+		Columns: []*plugin.Column{
+			{Name: "organization", Type: pb.ColumnType_STRING, Transform: transform.FromQual("organization")},
```
cbruno10, line 31: Can you please make sure every column has a description?

github/table_github_audit_log.go
```diff
@@ -0,0 +1,146 @@
+package github
+
+import (
+	"context"
+	"time"
+
+	"github.com/google/go-github/v45/github"
+	pb "github.com/turbot/steampipe-plugin-sdk/v3/grpc/proto"
```
cbruno10, line 8:
```suggestion
	"github.com/turbot/steampipe-plugin-sdk/v3/grpc/proto"
```

I believe we should be able to get rid of the alias here, and then for references below, instead of `pb`, we could use `proto`.

github/table_github_audit_log.go
```diff
@@ -0,0 +1,146 @@
...
+			{Name: "organization", Type: pb.ColumnType_STRING, Transform: transform.FromQual("organization")},
+			{Name: "phrase", Type: pb.ColumnType_STRING, Transform: transform.FromQual("phrase"), Description: "The search phrase for your audit events."},
+			{Name: "include", Type: pb.ColumnType_STRING, Transform: transform.FromQual("include"), Description: "The event types to include: web, git, all."},
+
+			// Top columns.
+			{Name: "id", Type: pb.ColumnType_STRING, Description: "The id of the audit event.", Transform: transform.FromField("DocumentID")},
+			{Name: "created_at", Type: pb.ColumnType_TIMESTAMP, Description: "The timestamp of the audit event.", Transform: transform.FromField("CreatedAt").Transform(convertTimestamp)},
+			{Name: "action", Type: pb.ColumnType_STRING, Description: "The action performed."},
+			{Name: "actor", Type: pb.ColumnType_STRING, Description: "The GitHub user who performed the action."},
+			{Name: "actor_location_country_code", Type: pb.ColumnType_STRING, Description: "The country location of the actor at the moment of the action.", Transform: transform.FromField("ActorLocation.CountryCode")},
```
cbruno10, line 40: What other data is contained in `ActorLocation`? I see here you're just pulling out the country code, but is there anything else useful in it? If so, we could have this column instead be called `actor_location` and type JSON.

github/table_github_audit_log.go
```diff
@@ -0,0 +1,146 @@
...
+			Hydrate: tableGitHubAuditLogList,
+		},
+		Columns: []*plugin.Column{
+			{Name: "organization", Type: pb.ColumnType_STRING, Transform: transform.FromQual("organization")},
+			{Name: "phrase", Type: pb.ColumnType_STRING, Transform: transform.FromQual("phrase"), Description: "The search phrase for your audit events."},
+			{Name: "include", Type: pb.ColumnType_STRING, Transform: transform.FromQual("include"), Description: "The event types to include: web, git, all."},
+
+			// Top columns.
+			{Name: "id", Type: pb.ColumnType_STRING, Description: "The id of the audit event.", Transform: transform.FromField("DocumentID")},
+			{Name: "created_at", Type: pb.ColumnType_TIMESTAMP, Description: "The timestamp of the audit event.", Transform: transform.FromField("CreatedAt").Transform(convertTimestamp)},
```
cbruno10, line 37: What's the difference between the timestamp in `CreatedAt` and the `Timestamp` property (https://github.com/google/go-github/blob/fd22ee93382e50d36592170f63c6a08d7c35ed8d/github/orgs_audit_log.go#L84)?

github/table_github_audit_log.go
```diff
@@ -0,0 +1,146 @@
...
+				{Name: "organization", Require: plugin.Required},
+				{Name: "phrase", Require: plugin.Optional},
+				{Name: "include", Require: plugin.Optional},
+				{Name: "action", Require: plugin.Optional},
+				{Name: "actor", Require: plugin.Optional},
+				{Name: "created_at", Require: plugin.Optional, Operators: []string{">", ">=", "<", "<=", "="}},
+			},
+			Hydrate: tableGitHubAuditLogList,
+		},
+		Columns: []*plugin.Column{
```
cbruno10, line 30: We don't necessarily need (or even want to include) all of the properties as columns for an audit log entry, but looking at https://github.com/google/go-github/blob/fd22ee93382e50d36592170f63c6a08d7c35ed8d/github/orgs_audit_log.go#L34-L92, there appear to be a lot of columns.

Are there any key properties that you feel we're missing in this table? For instance, what about `Event` or `Events`, or `Name` and `Message`? I'm not sure exactly what's in these properties, but the number of columns here compared to the struct seem a bit light.

github/table_github_audit_log.go
```diff
@@ -0,0 +1,146 @@
...
+			{Name: "organization", Type: pb.ColumnType_STRING, Transform: transform.FromQual("organization")},
+			{Name: "phrase", Type: pb.ColumnType_STRING, Transform: transform.FromQual("phrase"), Description: "The search phrase for your audit events."},
+			{Name: "include", Type: pb.ColumnType_STRING, Transform: transform.FromQual("include"), Description: "The event types to include: web, git, all."},
+
+			// Top columns.
+			{Name: "id", Type: pb.ColumnType_STRING, Description: "The id of the audit event.", Transform: transform.FromField("DocumentID")},
```
cbruno10, line 36: What does a document ID look like? And how common is that terminology in GitHub's UI, docs, etc.?

When column names deviate from property names the source uses, e.g., using `id` instead of `document_id`, there's usually a clear reason behind it. So in this case, if most people refer to audit logs' document IDs, if we change that name here, it may confuse some users if they're expecting a column called `document_id`.

docs/tables/github_audit_log.md
```diff
@@ -0,0 +1,60 @@
+# Table: github_audit_log
+
+The `github_audit_log` table helps to find all audit events for an organization.
+
+**You must always specify the organization** in the where or join clause using the `organization` column. Additionally, you can filter the logs by using a search phrase (`phrase`), event types (`include`), and before/after a timestamp (`created_at`).
+
+## Examples
+
+### Get recent audit events for an organization
```
cbruno10, line 9: By default, does GitHub give back the most recent audit log entries?

docs/tables/github_audit_log.md
```diff
@@ -0,0 +1,60 @@
+# Table: github_audit_log
+
+The `github_audit_log` table helps to find all audit events for an organization.
+
```
cbruno10, line 4: From my understanding, this table only works for organizations that have GitHub Enterprise enabled. If that's the case, can you please add a note in this table doc mentioning that requirement?

cbruno10 (changes requested):
@japborst I've left some initial review comments/suggestions, if you have any questions please let me know. Thanks!

ngowraj:
@japborst @cbruno10

> Can you confirm pagination does work now? i.e. there are more than 100 results

Yes, it does. 🙏 


> Can you confirm these fields are not null any more? id, created_at and actor_location_country_code

actor_location_country_code is still null for all records. 

> select count(*) as cnt, actor
from github_audit_log
where organization = 'loomhq' 
and phrase='action:protected_branch.policy_override created:2022-06-28'
group by actor;

This query works great. Thanks a lot. This will be really useful. 


> select
  * 
from
  github_audit_log 
where
  organization = 'loomhq' 
  and created_at > (created_at - interval '2' day)
limit 10;

Works 🙏 

Having filtering capability via the `phrase` column makes the queries really fast 🚀  💨

ngowraj:
Thanks for the team's effort and time. Looking forward to having this part of the main branch so that I can start using it for querying audit logs 🪵

japborst (commented):
Addressed comments

japborst:
@ngowraj awesome to hear that it works! 🎉 

These examples should now also work _fast_:

**Quickly filter on actions**
```sql
select
  id,
  created_at,
  actor,
  action,
  data
from
  github_audit_log
where
  organization = 'my_org'
  and action IN ('repo.create', 'repo.destroy')
  and created_at = '2022-01-01';
```

**Quickly filter on actor**
```sql
select
  id,
  created_at,
  actor,
  action,
  data
from
  github_audit_log
where
  organization = 'my_org'
  and actor = 'some_user'
  and created_at > (created_at - interval '30' day)
order by created_at;
```

github/table_github_audit_log.go
cbruno10: @japborst that sounds good to me, let's see if there's any feedback from other enterprise users

cbruno10:
@japborst @ngowraj - Thank you both so much for this PR + the extensive testing efforts! We really appreciate all of the contributions and are excited to have the ability to query audit logs directly.

PR #167, opened 2022-06-19
Refactor all Get hydrates

japborst:
There's a lot of duplication when hydrating the data. Likely from copy-pasting over time.

The aim of this PR is to decrease that for "get" hydrates.

Refactoring the list hydrates are a bit trickier, so I started out with just the gets. Lists can be done in a separate PR.

- [x] Refactor get hydrates into a single function
- [x] Always log upon hydrating

### DIFF
--- a/github/utils.go
+++ b/github/utils.go
@@ -74,6 +74,56 @@ func connect(ctx context.Context, d *plugin.QueryData) *github.Client {
 	return conn
 }
 
+type GetGitHubDataFunc func(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData, client *github.Client) (interface{}, error)
+
+func getGitHubItem(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData, getDetailsFunc GetGitHubDataFunc) (interface{}, error) {
+	getDetails := func(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
+		client := connect(ctx, d)
+
+		return getDetailsFunc(ctx, d, h, client)
+	}
+
+	data, err := plugin.RetryHydrate(ctx, d, h, getDetails, &plugin.RetryConfig{ShouldRetryError: shouldRetryError})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return data, err
+}
+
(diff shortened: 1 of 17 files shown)


github/utils.go
```diff
@@ -74,6 +74,56 @@ func connect(ctx context.Context, d *plugin.QueryData) *github.Client {
...
+			streamList(ctx, d, t)
+		case interface{}:
+			d.StreamListItem(ctx, t)
+		}
+	}
+
+	return nil, nil
+}
+
+func streamList(ctx context.Context, d *plugin.QueryData, item []interface{}) {
```
japborst, line 114: It would be nice to have this built into the SDK, similar to `d.StreamListItem`.

misraved:
Great work @japborst !!!

I had an initial pass at the changes and they look good. I will have a detailed look at it and provide feedback 👍 .

The idea of creating `streamList` function sounds interesting. Could you also create an issue in the [SDK repo](https://github.com/turbot/steampipe-plugin-github/pull/167#discussion_r901092798) so that we could discuss more on it.

japborst:
Thanks, @misraved.

Created the issue here https://github.com/turbot/steampipe-plugin-sdk/issues/341.

japborst:
@misraved I noticed I missed a few use-cases. Should be complete now 👍

japborst:
Not stale. Needs a review.

japborst:
@cbruno10 what would you like to do with this PR?

cbruno10:
Hey @japborst , sorry for the long delay. It's still on our radar to review, but we have some other items we're focusing on right now, and due to its scope, I don't want to rush a review. I'd like to leave the PR open and review some time in the next few weeks.

misraved:
Not stale. It will be reviewed shortly.

japborst:
Not stale

japborst:
Hey, thanks for the reply!

Using the GraphQL API makes a lot of sense. I really think you can get a big performance boost out of it, by reducing the number of calls to hydrate the data.

cbruno10:
@japborst Definitely, early tests are promising, so hoping we can get better performance so large queries and our mods will work smoother (rate limits are killer).

PR #200, opened 2022-09-27
feat: add github_code_owner table from .CODEOWNERS file in a repository

aminvielledebatAtBedrock:
# Description 

Capture the content of .CODEOWNERS file (in root, or in .github or docs directories), and expose it in a new table

# How

Use the Github API `client.Repositories.GetContents`

# Example query results
<details>
  <summary>Results</summary>
  
```
select 
    row_number() over (order by pattern, users, teams)
    ,repository_full_name
    ,pattern
    ,users
    ,teams
    ,pre_comments
    ,line_comment
from github.github_code_owner where repository_full_name = 'my-orga/code-owners'
```

| row\_number | repository\_full\_name             | pattern          | users                           | teams                  | pre\_comments                     | line\_comment       |
| ----------- | ---------------------------------- | ---------------- | ------------------------------- | ---------------------- | --------------------------------- | ------------------- |
| 1           | amy-orga/code-owners | .cloud           | \["@super-user","@super-user"\] | <null>                 | \["# another comment","# hello"\] | line comment here   |
| 2           | amy-orga/code-owners | .cloud/terraform | \["@super-user"\]               | <null>                 | \["# another comment","# hello"\] | line comment here 2 |
| 3           | amy-orga/code-owners | .github          | \["@super-user"\]               | \["@my-orga/my-team"\] | \["# fooo Comment","# here"\]     |                     |
| 4           | amy-orga/code-owners | README           | \["@super-user"\]               | <null>                 | \["# fooo Comment","# here"\]     |

</details>

# Dod

- [x] ~~Unit tests are written and everything is fine~~ (don't find any test in this repo :cry:)
- [x] update documentation
- [x] test it in a production environnement : Yes with more other 1000 repositories

### DIFF
--- a/docs/tables/github_code_owner.md
+++ b/docs/tables/github_code_owner.md
@@ -1,0 +1,90 @@
+# Table: github_code_owner
+
+All of your code owners rules defined in your repository, in CODEOWNERS files.
+
+The `github_code_owner` table can be used to query information about **ANY** repository, and **you must specify which repository** in the where or join clause (`where repository_full_name=`, `join github_code_owner on repository_full_name=`).
+
+## Examples
+
+### Get All your CodeOwners rules about a specific repository
+
+```sql
+select
+  line,
+  repository_full_name,
+  pattern,
+  users,
+  teams,
+  pre_comments,
+  line_comment 
+from
+  github_code_owner 
+where
+  repository_full_name = 'github/docs' 
+order by
+  line asc
+```
+
+**CODEOWNERS file content**
+
+```
+# Order is important. The LAST matching pattern has the MOST precedence.
+# gitignore style patterns are used, not globs.
+# https://docs.github.com/articles/about-codeowners
+# https://git-scm.com/docs/gitignore
+
+# Engineering
+*.js @github/docs-engineering
+*.ts @github/docs-engineering
+*.tsx @github/docs-engineering
+/.github/ @github/docs-engineering
+/script/ @github/docs-engineering
+/includes/ @github/docs-engineering
+/lib/search/popular-pages.json @github/docs-engineering
+Dockerfile @github/docs-engineering
+package-lock.json @github/docs-engineering
+package.json @github/docs-engineering
+
+# Localization
+/.github/actions-scripts/msft-create-translation-batch-pr.js @github/docs-engineering
+/.github/workflows/msft-create-translation-batch-pr.yml @github/docs-engineering
+/translations/ @Octomerger
+
+# Site Policy
+/content/site-policy/ @github/site-policy-admins
+
+# Content strategy
+/contributing/content-markup-reference.md @github/docs-content-strategy
+/contributing/content-style-guide.md @github/docs-content-strategy
+/contributing/content-model.md @github/docs-content-strategy
+/contributing/content-style-guide.md @github/docs-content-strategy
+/contributing/content-templates.md @github/docs-content-strategy
+
+# Requires review of #actions-oidc-integration, docs-engineering/issues/1506
+content/actions/deployment/security-hardening-your-deployments/** @github/oidc
+```
+
+**ResultSet**
+
+| line | repository\_full\_name | pattern                                                             | users             | teams                               | pre\_comments                                                                      | line\_comment |
+| ---- | ---------------------- | ------------------------------------------------------------------- | ----------------- | ----------------------------------- | ---------------------------------------------------------------------------------- | ------------- |
+| 7    | github/docs            | \*.js                                                               | <null>            | \["@github/docs-engineering"\]      | \["# Engineering"\]                                                                |               |
+| 8    | github/docs            | \*.ts                                                               | <null>            | \["@github/docs-engineering"\]      | \["# Engineering"\]                                                                |               |
+| 9    | github/docs            | \*.tsx                                                              | <null>            | \["@github/docs-engineering"\]      | \["# Engineering"\]                                                                |               |
+| 10   | github/docs            | /.github/                                                           | <null>            | \["@github/docs-engineering"\]      | \["# Engineering"\]                                                                |               |
+| 11   | github/docs            | /script/                                                            | <null>            | \["@github/docs-engineering"\]      | \["# Engineering"\]                                                                |               |
+| 12   | github/docs            | /includes/                                                          | <null>            | \["@github/docs-engineering"\]      | \["# Engineering"\]                                                                |               |
+| 13   | github/docs            | /lib/search/popular-pages.json                                      | <null>            | \["@github/docs-engineering"\]      | \["# Engineering"\]                                                                |               |
+| 14   | github/docs            | Dockerfile                                                          | <null>            | \["@github/docs-engineering"\]      | \["# Engineering"\]                                                                |               |
+| 15   | github/docs            | package-lock.json                                                   | <null>            | \["@github/docs-engineering"\]      | \["# Engineering"\]                                                                |               |
+| 16   | github/docs            | package.json                                                        | <null>            | \["@github/docs-engineering"\]      | \["# Engineering"\]                                                                |               |
+| 19   | github/docs            | /.github/actions-scripts/msft-create-translation-batch-pr.js        | <null>            | \["@github/docs-engineering"\]      | \["# Localization"\]                                                               |               |
+| 20   | github/docs            | /.github/workflows/msft-create-translation-batch-pr.yml             | <null>            | \["@github/docs-engineering"\]      | \["# Localization"\]                                                               |               |
+| 21   | github/docs            | /translations/                                                      | \["@Octomerger"\] | <null>                              | \["# Localization"\]                                                               |               |
+| 24   | github/docs            | /content/site-policy/                                               | <null>            | \["@github/site-policy-admins"\]    | \["# Site Policy"\]                                                                |               |
+| 27   | github/docs            | /contributing/content-markup-reference.md                           | <null>            | \["@github/docs-content-strategy"\] | \["# Content strategy"\]                                                           |               |
+| 28   | github/docs            | /contributing/content-style-guide.md                                | <null>            | \["@github/docs-content-strategy"\] | \["# Content strategy"\]                                                           |               |
+| 29   | github/docs            | /contributing/content-model.md                                      | <null>            | \["@github/docs-content-strategy"\] | \["# Content strategy"\]                                                           |               |
+| 30   | github/docs            | /contributing/content-style-guide.md                                | <null>            | \["@github/docs-content-strategy"\] | \["# Content strategy"\]                                                           |               |
+| 31   | github/docs            | /contributing/content-templates.md                                  | <null>            | \["@github/docs-content-strategy"\] | \["# Content strategy"\]                                                           |               |
+| 34   | github/docs            | content/actions/deployment/security-hardening-your-deployments/\*\* | <null>            | \["@github/oidc"\]                  | \["# Requires review of #actions-oidc-integration, docs-engineering/issues/1506"\] |
--- a/github/table_github_code_owner.go
+++ b/github/table_github_code_owner.go
@@ -1,0 +1,171 @@
+package github
+
+import (
+	"context"
+	"fmt"
+	"github.com/google/go-github/v45/github"
+	"github.com/turbot/steampipe-plugin-sdk/v4/plugin/transform"
+	"strings"
+
+	"github.com/turbot/steampipe-plugin-sdk/v4/grpc/proto"
+	"github.com/turbot/steampipe-plugin-sdk/v4/plugin"
+)
+
+//// TABLE DEFINITION
+
+func gitHubCodeOwnerColumns() []*plugin.Column {
+	return []*plugin.Column{
+		// Top columns
+		{Name: "repository_full_name", Type: proto.ColumnType_STRING, Description: "The full name of the repository, including the owner and repo name."},
+		// Other columns
+		{Name: "line", Type: proto.ColumnType_INT, Description: "The rule's line number in the CODEOWNERS file.", Transform: transform.FromField("LineNumber")},
+		{Name: "pattern", Type: proto.ColumnType_STRING, Description: "The pattern used to identify what code a team, or an individual is responsible for"},
+		{Name: "users", Type: proto.ColumnType_JSON, Description: "Users responsible for code in the repo"},
+		{Name: "teams", Type: proto.ColumnType_JSON, Description: "Teams responsible for code in the repo"},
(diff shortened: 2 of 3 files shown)


cbruno10:
Hey @aminvielledebatAtBedrock , thanks for working on this PR!

I think adding a way to get information from a repo's CODEOWNERS makes sense. GitHub treats the CODEOWNERS file as a key file, so I'm thinking maybe this could be a new table entirely, instead of being hidden among the other columns in a repo. If it were a table, it'd also be easier to query, with one line in the file per row instead of in a large JSON value.

I think a suitable table name would be `github_code_owner` and for columns:
- Instead of `path`, should that column be called `pattern`? In the [CODEOWNERS syntax](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax) doc, they refer to each rule as having a path and owners, e.g.,
```
*.txt @octo-org/octocats
```
- Since teams can also be set as owners, maybe we could separate out users, e.g., `@cbruno10`, `user@test.com`, into a `users` column and teams, e.g., `@octo-org/octocats`, into a `teams` column (both are list of strings).
- Each rule can also have a comment (single or multi-line), so could we include those in a few comments:
  - `{Name: "pre_comments", Type: proto.ColumnType_JSON, Description: "Specifies the comments added above a key."}`
  - `{Name: "head_comment", Type: proto.ColumnType_STRING, Description: "Specifies the comment in the lines preceding the node and not separated by an empty line."}`
  - `{Name: "line_comment", Type: proto.ColumnType_STRING, Description: "Specifies the comment at the end of the line where the node is in."}`
  - `{Name: "foot_comment", Type: proto.ColumnType_STRING, Description: "Specifies the comment following the node and before empty lines."}`
- I think order is important, as the last matching pattern takes the most precedence, so could we also add a `line` column, which is an int and represents which line number it's on (I'm not sure if a rule can traverse multiple lines though)?

If you have any questions, please let me know, thanks!

aminvielledebatAtBedrock:
Good point for the new table. I will add it !

About comments, what's the difference between head_coment and pre_comment for you ? IMO, it's the same thing :cry: 
And, how do you want to manage this kind of comments ? 

```
# In this example, @octocat owns any file in the `/apps`
# directory in the root of your repository except for the `/apps/github`
# subdirectory, as its owners are left empty.
/apps/ @octocat
/apps/github
```
Do you think `head_comment` and `pre_comment` should be replicated for pattern `/apps/` and `/app/github`.

Last point about `foot_comment` : i don't see any foot comment in the documentation. Do you think it's relevant ? 
Same question : do we have to replicate this comment for each pattern ? 

Thanks for your support !

cbruno10:
@aminvielledebatAtBedrock - I agree with your comments about the `comment` columns, so I think we can probably simplify to just `pre_comments` and `line_comment` for now.

Also, for the example you pasted, I believe yes, the values for the `pre_comments` column for the `/apps/` and `/app/github` rows should be the same.

aminvielledebatAtBedrock:
:pray: thank you so much for your responses @cbruno10 !
I updated my code with the following columns : 

``` go
func gitHubCodeOwnerColumns() []*plugin.Column {
	return []*plugin.Column{
		// Top columns
		{Name: "full_name", Type: proto.ColumnType_STRING, Description: "The full name of the repository, including the owner and repo name."},
		// Other columns
		{Name: "pattern", Type: proto.ColumnType_STRING, Description: "The pattern used to identify what code a team, or an individual is responsible for"},
		{Name: "users", Type: proto.ColumnType_JSON, Description: "Users responsible for code in the repo", Transform: transform.FromField("Users")},
		{Name: "teams", Type: proto.ColumnType_JSON, Description: "Teams responsible for code in the repo", Transform: transform.FromField("Teams")},
		{Name: "comments", Type: proto.ColumnType_JSON, Description: "Specifies the comments added above a key.", Transform: transform.FromField("Comments")},
	}
}
```

I also removed line comments because i don't see any example in the documentation, or in `.gitignore` files (CODEOWNERS have the same syntax)

If everything looks good to you, I can add the missing documentation.

cbruno10:
@aminvielledebatAtBedrock Thanks for the quick changes!

I think I'd still insist on having the `pre_comments` and `line_comment` columns. Though there aren't a lot of examples out there, GitHub announced support for line columns in this [blog post](https://github.blog/changelog/2022-02-17-codeowners-improvements-syntax-errors-preview-of-who-will-be-requested-and-more/#comment-on-the-same-line).

Also, can you please rename `full_name` to `repository_full_name`, as I believe all of our tables use that column name, so this table should be consistent with them.

docs/tables/github_code_owner.md
```diff
@@ -0,0 +1,44 @@
...
+All of your code owners rules defined in your repository, in CODEOWNERS files.
+
+The `github_code_owner` table can be used to query information about **ANY** repository, and **you must specify which repository** in the where or join clause (`where repository_full_name=`, `join github_code_owner on repository_full_name=`).
+
+## Examples
+
+### Get All your CodeOwners rules about a specific repository
+
+```sql
+select
```
cbruno10, line 12: @aminvielledebatAtBedrock Can you please format this query using https://www.freeformatter.com/sql-formatter.html? There's more information about our doc example standards in https://steampipe.io/docs/develop/table-docs-standards#style-conventions-1.

docs/tables/github_code_owner.md
```diff
@@ -0,0 +1,44 @@
...
+```sql
+select 
+    row_number() over (order by pattern, users, teams)
+    ,repository_full_name
+    ,pattern
+    ,users
+    ,teams
+    ,pre_comments
+    ,line_comment
+from github.github_code_owner where repository_full_name = 'my-orga/code-owners'
```
cbruno10, line 20:
```suggestion
from github_code_owner where repository_full_name = 'my-orga/code-owners'
```

For examples, we usually don't include the connection name, just the table name is sufficient (you can run queries like this too).

github/table_github_code_owner.go
```diff
@@ -0,0 +1,157 @@
...
+	"strings"
+
+	"github.com/turbot/steampipe-plugin-sdk/v4/grpc/proto"
+	"github.com/turbot/steampipe-plugin-sdk/v4/plugin"
+)
+
+//// TABLE DEFINITION
+
+func gitHubCodeOwnerColumns() []*plugin.Column {
+	return []*plugin.Column{
```
cbruno10, line 15: In the columns, can you please add a `line` column that's the line number the entry is found on? Then we can use this column directly instead of needing the `row_number()` function as well in the example.

github/table_github_code_owner.go
```diff
@@ -0,0 +1,157 @@
...
+		LineComment        string
+	}
+	getCodeOwners := func(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
+		var fileContent *github.RepositoryContent
+		var err error
+
+		client := connect(ctx, d)
+		opt := &github.RepositoryContentGetOptions{}
+		// stop on the first found CODEOWNERS file
+		var paths = []string{"CODEOWNERS", ".github/CODEOWNERS", "docs/CODEOWNERS"}
```
cbruno10, line 71: Can a repo have multiple CODEOWNERS files (even in GitHub throws up warnings in the UI)? If so, which one is used by the table (looks like it's the order defined in the array)? Is there a precedence that GitHub has on which CODEOWNERS file it will use?

cbruno10 (changes requested):
@aminvielledebatAtBedrock Please see additional suggestions/questions, thanks!

docs/tables/github_code_owner.md
aminvielledebatAtBedrock: Thanks for the links !

docs/tables/github_code_owner.md
aminvielledebatAtBedrock: Sorry about that, you're right. Fixed

github/table_github_code_owner.go
aminvielledebatAtBedrock: I add this column, but it leads to a non consecutive sequence if we point on the file line number (because we aggregate comments on a rule line).

Are you agree with that ?

github/table_github_code_owner.go
aminvielledebatAtBedrock: I did some tests on a repo : there is no composition between files (some rules from a file, and others from another) and there is a precedence :
- `.github/CODEOWNERS`
- `CODEOWNERS`
- `docs/CODEOWNERS`

I changed my code to respect this order.

cbruno10: Thanks for checking @aminvielledebatAtBedrock , can you please also include a comment that has a link where it specifies this for future reference?

github/table_github_code_owner.go
cbruno10: Yes, that's ok, even if the values of the `line` column across rows become non-consecutive, that's ok. I believe it's still useful so you can tell which rules are in which position in the file, which affects which rules take precedence.

github/table_github_code_owner.go
aminvielledebatAtBedrock: Comment added ! 
Thanks for the review !

github/table_github_code_owner.go
aminvielledebatAtBedrock: Good ! So it's done @cbruno10 :-)

docs/tables/github_code_owner.md
```diff
@@ -0,0 +1,52 @@
...
+select
+  line,
+  repository_full_name,
+  pattern,
+  users,
+  teams,
+  pre_comments,
+  line_comment 
+from
+  github.github_code_owner
```
cbruno10, line 21:
```suggestion
  github_code_owner 
```

github/table_github_code_owner.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	getCodeOwners := func(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
+		var fileContent *github.RepositoryContent
+		var err error
+
+		client := connect(ctx, d)
+		opt := &github.RepositoryContentGetOptions{}
+		// stop on the first found CODEOWNERS file.
+		// NOTE : a repository can have multiple CODEOWNERS files, even if it's invalid
+		// In that case, GitHub uses precedence over these files in the following order : .github/CODEOWNERS, CODEOWNERS, docs/CODEOWNERS
```
cbruno10, line 76: Can you please also include a link to the documentation where you found this precedence?

github/table_github_code_owner.go
```diff
@@ -0,0 +1,166 @@
...
+		var paths = []string{".github/CODEOWNERS", "CODEOWNERS", "docs/CODEOWNERS"}
+		for _, path := range paths {
+			fileContent, _, _, err = client.Repositories.GetContents(ctx, owner, repoName, path, opt)
+			if err == nil {
+				break
+			}
+		}
+
+		// no CODEOWNERS file
+		if err != nil {
```
cbruno10, line 86: Instead of handling all errors as the CODEOWNERS file not being present anywhere, in the `for loop` above, is there a way to instead:
- First check if the file being checked exists
  - If it does exist, try to get its contents
    - If we get an error, return that error immediately
  - If it doesn't exist, move onto the next file
- After the entire `for` loop, if we didn't find any CODEOWNERS files, return `[]CodeOwnerRuleResponse{}, nil`

My thought around this is that if there are other errors that API returns, I believe we should handle/return them so the users knows if something else is wrong, e.g., not enough access.

docs/tables/github_code_owner.md
```diff
@@ -0,0 +1,52 @@
...
+  repository_full_name,
+  pattern,
+  users,
+  teams,
+  pre_comments,
+  line_comment 
+from
+  github.github_code_owner 
+where
+  repository_full_name = 'my-orga/code-owners'
```
cbruno10, line 23:
```suggestion
  repository_full_name = 'github/docs' 
```

cbruno10: I think linking to a real example instead of having users search out repos with a CODEOWNERS file will make it easier for users to use this table and understand's returned

github/table_github_code_owner.go
```diff
@@ -0,0 +1,166 @@
...
+		{Name: "users", Type: proto.ColumnType_JSON, Description: "Users responsible for code in the repo"},
+		{Name: "teams", Type: proto.ColumnType_JSON, Description: "Teams responsible for code in the repo"},
+		{Name: "pre_comments", Type: proto.ColumnType_JSON, Description: "Specifies the comments added above a key."},
+		{Name: "line_comment", Type: proto.ColumnType_STRING, Description: "Specifies the comment following the node and before empty lines."},
+	}
+}
+
+//// TABLE DEFINITION
+
+func tableGithubCodeOwner() *plugin.Table {
```
cbruno10, line 31:
```suggestion
func tableGitHubCodeOwner() *plugin.Table {
```

cbruno10 (changes requested):
@aminvielledebatAtBedrock Thanks again for the quick updates! I left one more round of suggestions, but I believe after these are addressed, the PR will be ready for merge. Thanks!

github/table_github_code_owner.go
aminvielledebatAtBedrock: I don't find any documentation about that :'(.
The precedence, and github behaviour is only based on my tests on a personnal repository.

I looked on github.blog with no success

cbruno10: Thanks for checking @aminvielledebatAtBedrock , I couldn't find any docs either, but if they are based off of your test results, that's good enough for me.

github/table_github_code_owner.go
aminvielledebatAtBedrock: Done @cbruno10 !

Github API doesn't provide a way to test if a file exists.

The only way i found : 
- use `client.Repositories.GetContents` on each path.
- If err is not nil and `err.(*github.ErrorResponse).Response.StatusCode ` is different than 404, return that error immediatly with `err.(*github.ErrorResponse).Response.Status`.
- if `StatusCode` is equal to 404, move onto the next file
- if err is nil, use the file content

I think, the logic is similar.

cbruno10: Sounds good to me, thanks!

cbruno10 (approved):

cbruno10:
@aminvielledebatAtBedrock Thanks again so much for submitting this PR and putting up with all of my reviews 😅 

We're doing another GitHub plugin release later this week, so this table will be included in it.

aminvielledebatAtBedrock:
@cbruno10 Thanks again for your support and your patience :laughing: 
It's my first PR on steampipe and i'm very happy to work with you :-)

For the github release : i'm working on another table if you're interested : `github_repository_content`. To be continued

PR #205, opened 2022-09-30
Retry secondary ratelimits

TheJokersThief:
# What 

* Add AbuseRateLimit to collection of errors to be retried

# Why

Github has this concept of secondary or "abuse" ratelimits where if a number of queries are made in succession, you are ratelimited based on your CPU usage, and not by the normal ratelimit. They recommend you wait a moment and try again if it's a legitimate query.

More info on secondary rate limit: https://docs.github.com/en/rest/overview/resources-in-the-rest-api#secondary-rate-limits

### DIFF
--- a/github/errors.go
+++ b/github/errors.go
@@ -13,6 +13,11 @@ func shouldRetryError(err error) bool {
 		log.Printf("[WARN] Received Rate Limit Error")
 		return true
 	}
+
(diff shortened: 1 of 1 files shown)


cbruno10 (approved):

cbruno10:
Nice catch @TheJokersThief , thanks for submitting this PR!

PR #221, opened 2022-12-07
Fix broken plugin in README

octo-topi:
Plugin install fails because plugin name is incorrect

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 Install the plugin with [Steampipe](https://steampipe.io/downloads):
 
 ```shell
-steampipe plugin install github
+steampipe plugin install turbot/github
 ```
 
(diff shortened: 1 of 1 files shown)


misraved:
Welcome to Steampipe @octo-topi and thanks for raising the PR 👍.

Typically, `steampipe plugin install github` should not fail. COuld you please share the error that you get when you run the command.

Are you getting similar errors when you run `steampipe plugin install aws` as well?

cbruno10:
@octo-topi For most plugin installations, you shouldn't need to give the organization/owner name, like Turbot. Do you get an error when trying to run the command in the current README? If so, can you please share it? Thanks!

cbruno10:
@octo-topi No problem, if you encounter it again, please let us know and we're happy to help debug more!

PR #237, opened 2023-05-02
feat: updated github_license to utilise GraphQL

graza-io:


docs/tables/github_license.md
```diff
@@ -39,3 +39,48 @@ group by
 order by
   num_repos desc;
 ```
+
+### View conditions for a specific license
```
cbruno10, line 43: Great new examples!

github/table_github_license.go
```diff
@@ -27,53 +26,67 @@ func tableGitHubLicense() *plugin.Table {
 		Columns: []*plugin.Column{
 
 			// Top columns
-			{Name: "spdx_id", Description: "The Software Package Data Exchange (SPDX) id of the license.", Type: proto.ColumnType_STRING, Transform: transform.FromField("SPDXID")},
+			{Name: "spdx_id", Description: "The Software Package Data Exchange (SPDX) id of the license.", Type: proto.ColumnType_STRING, Transform: transform.FromField("SpdxId")},
 			{Name: "name", Description: "The name of the license.", Type: proto.ColumnType_STRING},
-			{Name: "html_url", Description: "The HTML URL of the license.", Type: proto.ColumnType_STRING, Hydrate: tableGitHubLicenseGetData},
+			{Name: "html_url", Description: "The HTML URL of the license.", Type: proto.ColumnType_STRING, Transform: transform.FromField("Url")},
```
cbruno10, line 31: To stay in line with GitHub's API (like we do with other columns), should we rename this column to `url` instead?

github/table_github_license.go
```diff
@@ -27,53 +26,67 @@ func tableGitHubLicense() *plugin.Table {
...
-			licenses: licenses,
-			resp:     resp,
-		}, err
-	}
+var getLicenseQuery struct {
+	License license `graphql:"license(key: $key)"`
+}
 
-	listPageResponse, err := retryHydrate(ctx, d, h, listPage)
+// // LIST FUNCTION
```
cbruno10, line 77:
```suggestion
// LIST FUNCTION
```

Do we need two sets of `//`'s?

cbruno10 (changes requested):
@graza-io Please see review comments, thanks!

github/table_github_license.go
graza-io: External urls appear to be named html_url in other tables also, I'd recommend leaving as such for now and reviewing as we change over other tables to v4 if this is something we can make a standard approach on?

github/table_github_license.go
graza-io: Ha, originally it was 4 together (e.g. `//// HYDRATE FUNCTIONS`, my IDE appears to have inserted a space in between them - I don't mind using 4 or 2 is there a standard?

cbruno10: If other tables use 4 `/`'s, then I'm ok to keep that style

graza-io: Ok, I'll make adjustment to remove the extraneous space (& dig into my IDE config to stop the behaviour of inserting spaces there...) shortly.

cbruno10 (approved):

PR #239, opened 2023-05-03
Updated github_tag to utilise GraphQL

graza-io:


github/table_github_tag.go
```diff
@@ -25,69 +23,162 @@ func tableGitHubTag(ctx context.Context) *plugin.Table {
...
+			{Name: "date", Type: proto.ColumnType_TIMESTAMP, Transform: transform.FromField("Date").NullIfZero(), Description: "Date the tag was created."},
+			{Name: "tagger_name", Type: proto.ColumnType_STRING, Description: "Name of user whom created the tag."},
+			{Name: "tag_message", Type: proto.ColumnType_STRING, Description: "Message associated with the tag."},
+			{Name: "commit_sha", Type: proto.ColumnType_STRING, Description: "Commit SHA the tag refers to."},
+			{Name: "commit_short_sha", Type: proto.ColumnType_STRING, Description: "Commit short SHA the tag refers to."},
+			{Name: "commit_date", Type: proto.ColumnType_TIMESTAMP, Transform: transform.FromField("CommitDate").NullIfZero(), Description: "Date the commit referenced by the tag was made."},
+			{Name: "commit_message", Type: proto.ColumnType_STRING, Description: "Commit message of the commit the tag is referencing."},
+			{Name: "commit_author", Type: proto.ColumnType_STRING, Description: "Name of the author for the commit the tag is referencing."},
+			{Name: "commit_url", Type: proto.ColumnType_STRING, Transform: transform.FromField("CommitUrl"), Description: "Commit URL the tag refers to."},
```
cbruno10, line 34: Do we need the `transform.FromField`? Does the default transform the plugin use fail to calculate the column name?

cbruno10 (changes requested):
@graza-io Left 1 minor comment, please have a look. Thanks!

github/table_github_tag.go
graza-io: It pulled `null` initially, so I added the FromField transform to alleviate this.

cbruno10: What's the default transformation of the plugin? Would you expect that to work for the field name we get back?

graza-io: `DefaultTransform: transform.FromGo()`

I haven't tested extensively but it seemingly prefers capitalisation for terms like `WebURL` -> `web_url`

cbruno10: Oh yeah that's correct, `FromGo` capitalizes some common initialisms (https://github.com/turbot/go-kit/blob/d0698ef1044e8442a58b50c9a24cfab53ede6895/helpers/lint_name.go#L79). Transforms are OK then

PR #292, opened 2023-06-26
fix: resolves issue of no perms for collaborators on repo tables, adds affiliation filter to repo_collaborator

graza-io:


github/table_github_repository_collaborator.go
```diff
@@ -66,6 +95,9 @@ func tableGitHubRepositoryCollaboratorList(ctx context.Context, d *plugin.QueryD
 		_, err := plugin.RetryHydrate(ctx, d, h, listPage, retryConfig())
 		plugin.Logger(ctx).Debug(rateLimitLogString("github_repository_collaborator", &query.RateLimit))
 		if err != nil {
+			if strings.Contains(err.Error(), "You do not have permission to view repository collaborators") {
```
cbruno10, line 98: Do we need this separate log statement if we have the `Error` log statement below, which will show up in the plugin logs?

cbruno10 (changes requested):
@graza-io Please see review, thanks!

cbruno10 (approved):

PR #299, opened 2023-07-17
Add labels_src/labels columns to issue/pr tables. Closes #298

### LINKED ISSUE #298: Add labels/tags to issue/PR tables.
**Context**
@judell mentioned that since we migrated to the GraphQL SDK that he is missing the `tags` from the issue tables, upon inspection these appear to just be parsed variants of the `labels` returned by the APIs.

However, it appears that we omitted the labels & instead included a `labels_total_count` field on the following tables due to labels being a paginated nested resource:
- `github_my_issue`
- `github_issue`
- `github_pull_request`

**Describe the solution you'd like**
Since the expected number of labels on a typical issue / PR would be less than a single API page (100), it makes sense to obtain the first page of labels in the aforementioned tables as a nested item (JSON Column), this approach may mean that we have edge cases where an issue or PR which has in-excess of 100 labels may be missing the data from the column on the table.

**Alternative Solution**
Implement as a separate table that can be joined onto when labels are required (`github_issue_label`, `github_pull_request_label`) - this approach would ensure we don't miss data, but to obtain labels for a collection of issues/PRs would incur `n` x rate limit credit usage where `n` is number of issues/PRs.

**Additional context**
Both are valid approaches as per the [Design Principals](https://github.com/turbot/steampipe-utils/wiki/GraphQL-Plugin-Design-Principals) - @cbruno10 primarily looking for your preferred approach and I can implement that.

graza-io:


github/table_github_issue.go
```diff
@@ -52,6 +52,8 @@ func sharedIssueColumns() []*plugin.Column {
 		{Name: "assignees_total_count", Type: proto.ColumnType_INT, Transform: transform.FromField("Assignees.TotalCount", "Node.Assignees.TotalCount"), Description: "Count of assignees on the issue."},
 		{Name: "comments_total_count", Type: proto.ColumnType_INT, Transform: transform.FromField("Comments.TotalCount", "Node.Comments.TotalCount"), Description: "Count of comments on the issue."},
 		{Name: "labels_total_count", Type: proto.ColumnType_INT, Transform: transform.FromField("Labels.TotalCount", "Node.Labels.TotalCount"), Description: "Count of labels on the issue."},
+		{Name: "labels_src", Type: proto.ColumnType_JSON, Transform: transform.FromField("Labels.Nodes", "Node.Labels.Nodes"), Description: "The first 100 labels associated to the issue."},
+		{Name: "labels", Type: proto.ColumnType_JSON, Description: "A map of label names associated with this issue, in Steampipe standard format.", Transform: transform.From(getIssueLabels)},
```
cbruno10, line 56:
```suggestion
		{Name: "labels", Type: proto.ColumnType_JSON, Description: "A map of labels for the issue.", Transform: transform.From(getIssueLabels)},
```

I think a shorter description is OK, as "Steampipe standard format" isn't that specific. This also matches the description we use in other plugins.

cbruno10 (changes requested):
@graza-io Please see comments, thanks!

cbruno10 (approved):

PR #330, opened 2023-09-11
chore: minor updates to deps

graza-io:


graza-io:
These were tested working, reason for not upgrading `github.com/argonsecurity/pipeline-parser` to latest was that it appears to be designed for `go 1.21` which I'm not sure we're upgrading plugins to yet:

```
go build -o ~/.steampipe/plugins/hub.steampipe.io/plugins/turbot/github@latest/steampipe-plugin-github.plugin *.go
# github.com/quic-go/quic-go/internal/qtls
../../../go/pkg/mod/github.com/quic-go/quic-go@v0.32.0/internal/qtls/go121.go:5:13: cannot use "The version of quic-go you're using can't be built on Go 1.21 yet. For more details, please see https://github.com/quic-go/quic-go/wiki/quic-go-and-Go-versions." (untyped string constant "The version of quic-go you're using can't be built on Go 1.21 yet. F...) as int value in variable declaration
make: *** [install] Error 1
```

graza-io:
Also just noticed we have Dependabot PRs for some of these...

cbruno10:
@graza-io Can we use the Dependabot PRs for any packages covered in both sets of PRs? Any not caught by Dependabot we can then update in this PR.

graza-io:
The following Dependabot PRs are captured here & tested but should just use their own PR:
- https://github.com/turbot/steampipe-plugin-github/pull/328
- https://github.com/turbot/steampipe-plugin-github/pull/327
  - Albeit this one will be replaced by v55 in this PR: https://github.com/turbot/steampipe-plugin-github/pull/332
- https://github.com/turbot/steampipe-plugin-github/pull/325

The following Dependabot PR is probably a good one to include but untested:
- https://github.com/turbot/steampipe-plugin-github/pull/326

This [one](https://github.com/turbot/steampipe-plugin-github/pull/324) can't be done right now as [latest would break build](https://github.com/turbot/steampipe-plugin-github/pull/330#issuecomment-1713369936), which is the main purpose of this PR, to update to a reasonable version in between current and latest, which contains some fixes.

graza-io:
@cbruno10 reworked so now only contains the update for `pipeline-parser`.

cbruno10:
@graza-io I think we could update to use Go 1.21 (the latest Steampipe plugin SDK will require it I believe), unless there's a larger issue that you know of. If we update to Go 1.21, would we be able to merge in https://github.com/turbot/steampipe-plugin-github/pull/324, or are there other breaking changes in argonsecurity/pipeline-parser v1.3.4?

graza-io:
Actually I think I read this wrong and that it requires being built with a version of go prior to 1.21.

Starting with `v1.0.0` of `pipeline-parser` they've taken a dependency on `quic` 
[go.mod @ v0.3.3](https://github.com/argonsecurity/pipeline-parser/blob/v0.3.3/go.mod)
[go.mod @ v1.0.0](https://github.com/argonsecurity/pipeline-parser/blob/v1.0.0/go.mod#L25-L29)

This seems to have [issues](https://github.com/quic-go/quic-go/wiki/quic-go-and-Go-versions) being built.
[Using go1.21 at v0.32.0 of go-quic](https://github.com/quic-go/quic-go/blob/v0.32.0/internal/qtls/go121.go) - as a dependency on [v.3.33.1 of req](https://github.com/imroc/req/blob/v3.33.1/go.mod)

Because my local `go` version is 1.21, I cannot compile the code.
```sh
❯ go version
go version go1.21.1 darwin/arm64
```

`go-quic` remediates the support in [v0.37.0](https://github.com/quic-go/quic-go/blob/v0.37.0/internal/qtls/go121.go)
`req` picks up the newer `go-quic` in [v3.38.0](https://github.com/imroc/req/blob/v3.38.0/go.mod#L8)

Options:
- Update `pipeline-parser` to `0.3.3` and watch for a new version which includes dependency upgrade of `req` >= `v3.38.0`, this will allow us to continue building locally with go1.21 but also to use go1.21 as plugin version.
- Update `pipeline-parser` to `1.3.4` and be limited to go1.20 for building locally and in plugin version until a newer release is available and we update to that.

cbruno10:
@graza-io Thanks for the details, my vote would be to update to `v0.3.3` then for now, since we will be upgrading the plugin to use the new Steampipe plugin SDK version in the upcoming weeks, which requires Go v1.21.

cbruno10 (approved):

PR #338, opened 2023-09-20
feat: repository table dynamic gql requests

graza-io:


graza-io:
Just a note that the `lint` failure is due to [limiting to g1.19](https://github.com/turbot/steampipe-workflows/blob/main/.github/workflows/golangci-lint.yml#L20) in the workflow.

cbruno10 (approved):

PR #342, opened 2023-09-28
feat: issue table dynamically adjust gql query based on req'd cols

graza-io:


cbruno10:
Hey @misterek , I'm actually reviewing this PR now and plan to release it later today. Once it's released, you should see this PR close.

github/models/issue.go
```diff
@@ -7,42 +7,42 @@ type Issue struct {
 	NodeId                  string                               `graphql:"nodeId: id" json:"node_id"`
 	Number                  int                                  `json:"number"`
 	ActiveLockReason        githubv4.LockReason                  `json:"active_lock_reason"`
-	Author                  Actor                                `json:"author"`
+	Author                  Actor                                `graphql:"author @include(if:$includeIssueAuthor)" json:"author"`
```
cbruno10, line 10: @graza-io How were columns in each set of issue/PR tables selected to be updated vs. leaving them as is?

graza-io: Started with moving over embedded items (including counts), then also went with adding those which required checks to user permissions as these seemed to add a noticeable overhead (not huge but a decent percentage of the total time.

Could continue and have all columns (except key ones) as dynamic but this pass I just wanted to focus on optimisations that were measurable.

cbruno10: Thanks @graza-io , that makes sense, I think starting with the column types you mentioned first makes sense.

cbruno10 (approved):

PR #353, opened 2023-10-21
Add github_repository_sbom table

lwakefield:
# Example query results
<details>
  <summary>Results</summary>
  
```sql
> select              
  repository_full_name,
  p->>'name' as package_name,
  p->>'versionInfo' as package_version,
  p->>'licenseConcluded' as package_license
from
  github_repository_sbom,
  jsonb_array_elements(packages) p
where
  repository_full_name = 'turbot/steampipe'
limit 10;
+----------------------+-------------------------------------------+-----------------+-----------------+
| repository_full_name | package_name                              | package_version | package_license |
+----------------------+-------------------------------------------+-----------------+-----------------+
| turbot/steampipe     | com.github.turbot/steampipe               |                 | <null>          |
| turbot/steampipe     | npm:@actions/core                         | 1.2.6           | MIT             |
| turbot/steampipe     | npm:@actions/github                       | 4.0.0           | MIT             |
| turbot/steampipe     | npm:@actions/http-client                  | 1.0.9           | MIT             |
| turbot/steampipe     | npm:@octokit/auth-token                   | 2.4.3           | MIT             |
| turbot/steampipe     | npm:@octokit/core                         | 3.2.1           | MIT             |
| turbot/steampipe     | npm:@octokit/endpoint                     | 6.0.9           | MIT             |
| turbot/steampipe     | npm:@octokit/graphql                      | 4.5.7           | MIT             |
| turbot/steampipe     | npm:@octokit/plugin-paginate-rest         | 2.6.0           | MIT             |
| turbot/steampipe     | npm:@octokit/plugin-rest-endpoint-methods | 4.2.1           | MIT             |
```
</details>

First time contributor! I'm looking for / expecting feedback on:
1. ~Is this the right way to model the data since it is focusing on the packages and not top level SBOM metadata? I can see a possibility where it might be preferred to move this to a `github_repository_sbom_packages` table instead, but unsure if that is the right move or not.~ - Pivoted based on feedback!
2. ~Pending 2 - there are some fields missing, I wanted to get feedback before adding them!~ - Addressed!

### DIFF
--- a/docs/tables/github_repository_sbom.md
+++ b/docs/tables/github_repository_sbom.md
@@ -1,0 +1,20 @@
+# Table: github_repository_sbom
+
+The `github_repository_sbom` table can be used to query information about packages listed in the SBOM of a repository.
+
+**You must specify which repository** in the where or join clause using the `repository_full_name` column.
+
+## Examples
+
+### List SBOM packages
+
+```sql
+select
+  name,
+  version,
+  license
+from
+  github_repository_sbom
+where
+  repository_full_name = 'turbot/steampipe';
+```
--- a/docs/tables/github_repository_sbom.md
+++ b/docs/tables/github_repository_sbom.md
@@ -10,11 +10,13 @@ The `github_repository_sbom` table can be used to query information about packag
 
 ```sql
 select
-  name,
-  version,
-  license
+  github_repository_sbom,
+  p->>'name' as package_name,
+  p->>'versionInfo' as package_version,
+  p->>'licenseConcluded' as package_license
 from
-  github_repository_sbom
+  github_repository_sbom,
+  jsonb_array_elements(packages) p
 where
   repository_full_name = 'turbot/steampipe';
 ```
--- a/github/table_github_repository_sbom.go
+++ b/github/table_github_repository_sbom.go
@@ -30,24 +30,53 @@ func tableGitHubRepositorySbom() *plugin.Table {
 				Description: "The full name of the repository (login/repo-name).",
 			},
 			{
+				Name:        "spdx_id",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("SPDXID"),
+				Description: "The SPDX identifier for the SPDX document.",
+			},
+			{
+				Name:        "spdx_version",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("SPDXVersion"),
+				Description: "The version of the SPDX specification that this document conforms to.",
+			},
+			{
+				Name:        "creation_info",
+				Type:        proto.ColumnType_JSON,
+				Transform:   transform.FromField("CreationInfo"),
+				Description: "The version of the SPDX specification that this document conforms to.",
+			},
+			{
 				Name:        "name",
 				Type:        proto.ColumnType_STRING,
 				Transform:   transform.FromField("Name"),
-				Description: "The name of the package.",
+				Description: "The name of the SPDX document.",
 			},
 			{
-				Name:        "version",
+				Name:        "data_license",
 				Type:        proto.ColumnType_STRING,
-				Transform:   transform.FromField("VersionInfo"),
-				Description: "Version info of the package.",
+				Transform:   transform.FromField("DataLicense"),
+				Description: "The license under which the SPDX document is licensed.",
 			},
 			{
-				Name:        "license",
+				Name:        "document_describes",
 				Type:        proto.ColumnType_STRING,
-				Transform:   transform.FromField("LicenseConcluded"),
-				Description: "License of the package.",
+				Transform:   transform.FromField("DocumentDescribes"),
+				Description: "The name of the repository that the SPDX document describes.",
 			},
-			// TODO: there are more fields to be added here!
+			{
+				Name:        "document_namespace",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("DocumentNamespace"),
+				Description: "The namespace for the SPDX document.",
+			},
+			{
+				Name:        "packages",
+				Type:        proto.ColumnType_JSON,
+				Transform:   transform.FromField("Packages"),
+				Description: "Array of packages in spdx format.",
+			},
 		},
 	}
 }
@@ -68,14 +97,7 @@ func tableGitHubRepositorySbomList(ctx context.Context, d *plugin.QueryData, h *
 		return nil, err
 	}
 
-	for _, i := range sbom.SBOM.Packages {
-		d.StreamListItem(ctx, i)
-
-		// Context can be cancelled due to manual cancellation or the limit has been hit
-		if d.RowsRemaining(ctx) == 0 {
-			return nil, nil
-		}
-	}
+	d.StreamListItem(ctx, sbom.SBOM)
 
 	return sbom, nil
 }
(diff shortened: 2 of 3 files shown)


cbruno10:
Hey @lwakefield , thanks for opening this PR!

I had a look at the [SBOM API](https://docs.github.com/en/free-pro-team@latest/rest/dependency-graph/sboms?apiVersion=2022-11-28), and I think including top level SBOM data is OK and having the `packages` in its own column. So there will be additional columns like `spdx_id` ,`spdx_version`, `creation_info`, `name`, data_license`, etc., along with `packages`, which is a JSONB column.

To make querying easier in examples, you can include an example like:
```sql
select
  s.name as spdx_name,
  p.name as package_name,
  p.version as package_version,
  p.license as package_license
from
  github_repository_sbom as s,
  jsonb_array_elements(packages) as p
where
  repository_full_name = 'turbot/steampipe';
```

Also, I did have a quick look at if this was possible in GitHub's GraphQL API, and it seems like right now it's only in [preview mode](https://docs.github.com/en/graphql/overview/schema-previews#access-to-a-repositorys-dependency-graph-preview), which means it's also subject to API changes, so I think implementing this with the v3 REST API right now makes sense, and we can look at migrating to the GraphQL API when it's more widely available.

lwakefield:
Thanks @cbruno10 - I pivoted based on your feedback!

github/table_github_repository_sbom.go
```diff
@@ -0,0 +1,103 @@
...
+			{
+				Name:        "data_license",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("DataLicense"),
+				Description: "The license under which the SPDX document is licensed.",
+			},
+			{
+				Name:        "document_describes",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("DocumentDescribes"),
```
cbruno10, line 65: Do we need the transform for this column (and others)? Usually we only need these when the default transformation (https://github.com/turbot/steampipe-plugin-github/blob/main/github/plugin.go#L18) fails, usually when there's a field name with acronyms, e.g., `SPDXID`. I think we may only need these explicit transforms for the `spdx_id` and maybe the `spdx_version` columns.

github/table_github_repository_sbom.go
```diff
@@ -0,0 +1,103 @@
...
+			},
+			{
+				Name:        "data_license",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("DataLicense"),
+				Description: "The license under which the SPDX document is licensed.",
+			},
+			{
+				Name:        "document_describes",
+				Type:        proto.ColumnType_STRING,
```
cbruno10, line 64: For this column, do we get get always a single string, or can we get back a list? Looking at https://docs.github.com/en/rest/dependency-graph/sboms?apiVersion=2022-11-28, the examples show the data as lists.

github/table_github_repository_sbom.go
```diff
@@ -0,0 +1,103 @@
...
+				Name:        "spdx_version",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("SPDXVersion"),
+				Description: "The version of the SPDX specification that this document conforms to.",
+			},
+			{
+				Name:        "creation_info",
+				Type:        proto.ColumnType_JSON,
+				Transform:   transform.FromField("CreationInfo"),
+				Description: "The version of the SPDX specification that this document conforms to.",
```
cbruno10, line 48: I think this was a copy and paste error from above

github/table_github_repository_sbom.go
```diff
@@ -0,0 +1,103 @@
...
+				Name:        "document_namespace",
+				Type:        proto.ColumnType_STRING,
+				Transform:   transform.FromField("DocumentNamespace"),
+				Description: "The namespace for the SPDX document.",
+			},
+			{
+				Name:        "packages",
+				Type:        proto.ColumnType_JSON,
+				Transform:   transform.FromField("Packages"),
+				Description: "Array of packages in spdx format.",
```
cbruno10, line 78:
```suggestion
				Description: "Array of packages in SPDX format.",
```

github/table_github_repository_sbom.go
```diff
@@ -0,0 +1,103 @@
...
+
+	"github.com/turbot/steampipe-plugin-sdk/v5/grpc/proto"
+	"github.com/turbot/steampipe-plugin-sdk/v5/plugin"
+	"github.com/turbot/steampipe-plugin-sdk/v5/plugin/transform"
+)
+
+func tableGitHubRepositorySbom() *plugin.Table {
+	return &plugin.Table{
+		Name:        "github_repository_sbom",
+		Description: "SBOM from a repository.",
```
cbruno10, line 14:
```suggestion
		Description: "Get the software bill of materials (SBOM) for a repository.",
```

Description partially taken from https://docs.github.com/en/rest/dependency-graph/sboms?apiVersion=2022-11-28.

docs/tables/github_repository_sbom.md
```diff
@@ -0,0 +1,22 @@
...
+
+**You must specify which repository** in the where or join clause using the `repository_full_name` column.
+
+## Examples
+
+### List SBOM packages
+
+```sql
+select
+  github_repository_sbom,
```
cbruno10, line 13: Does this example work? I don't think having the table name in the `select` statement would yield anything.

I'd suggest adding some basic SBOM columns like `spdx_id` and `spdx_version` instead.

docs/tables/github_repository_sbom.md
```diff
@@ -0,0 +1,22 @@
+# Table: github_repository_sbom
+
+The `github_repository_sbom` table can be used to query information about packages listed in the SBOM of a repository.
+
+**You must specify which repository** in the where or join clause using the `repository_full_name` column.
+
+## Examples
```
cbruno10, line 7: Are there any other examples you'd want to include based on how you'd use the table to view SBOM/package information? For instance, is there anything in `packages` you'd want to validate specifically?

github/table_github_repository_sbom.go
```diff
@@ -0,0 +1,103 @@
...
+
+import (
+	"context"
+
+	"github.com/turbot/steampipe-plugin-sdk/v5/grpc/proto"
+	"github.com/turbot/steampipe-plugin-sdk/v5/plugin"
+	"github.com/turbot/steampipe-plugin-sdk/v5/plugin/transform"
+)
+
+func tableGitHubRepositorySbom() *plugin.Table {
```
cbruno10, line 11: Can you please post what `select * from github_repository_sbom` produces?

docs/tables/github_repository_sbom.md
```diff
@@ -0,0 +1,22 @@
...
+
+### List SBOM packages
+
+```sql
+select
+  github_repository_sbom,
+  p->>'name' as package_name,
+  p->>'versionInfo' as package_version,
+  p->>'licenseConcluded' as package_license
```
cbruno10, line 16:
```suggestion
  p ->> 'name' as package_name,
  p ->> 'versionInfo' as package_version,
  p ->> 'licenseConcluded' as package_license
```

cbruno10 (changes requested):
@lwakefield Thanks for making the quick pivot!

I've left a few additional suggestions, can you please have a look?

madhushreeray30:
Re-raised as new PR

PR #449, opened 2024-08-22
added support for fine grained token

ritin0204:


ritin0204:
docs: https://github.blog/security/application-security/introducing-fine-grained-personal-access-tokens-for-github/

ParthaI:
Hi, @ritin0204, thanks for the contribution! 🎉

Could you please take a look at the following points?
- Could you update the PR description with an example query using the fine-grained token? Please include the results for tables that use both the REST API and GraphQL.
- Please update the [GraphQL client](https://github.com/turbot/steampipe-plugin-github/blob/main/github/utils.go#L194) to utilize the fine-grained token.
- Update the [plugin documentation](https://github.com/turbot/steampipe-plugin-github/blob/main/docs/index.md?plain=1#L83) and the [config file](https://github.com/turbot/steampipe-plugin-github/blob/main/config/github.spc#L8) to reflect these changes.

Thanks!

ParthaI:
Hey @ritin0204, sorry for the brevity of my previous review comment, as it may not have provided enough clarity to fully explain the feedback.

> can you provide me some resources to set it up and test it locally?

Could you please clarify what specific information or resources you need from us?

In my [previous comment](https://github.com/turbot/steampipe-plugin-github/pull/449#issuecomment-2304070259), I intended to highlight the following:

- In the GitHub Steampipe plugin, we use two types of clients:
   - [`connect()`](https://github.com/turbot/steampipe-plugin-github/blob/main/github/utils.go#L26) for the REST API (v3) client.
   - [`connectV4()`](https://github.com/turbot/steampipe-plugin-github/blob/main/github/utils.go#L146C6-L146C16) for the GraphQL API (v4) client.
- We restrict the types of tokens that can be used when creating clients for both [V3](https://github.com/turbot/steampipe-plugin-github/blob/main/github/utils.go#L78) and [V4](https://github.com/turbot/steampipe-plugin-github/blob/main/github/utils.go#L194).
- However, the changes made in this PR only address the `v3` client and not `v4`. 
   - As a result, tables like `github_search_user`, `github_branch`, `github_tag`, etc... that rely on the `v4` client will encounter errors with these changes if the fine-grained access token is used to query the tables.

> As the change was just to add support for fine grained token same as other tokens so i haven't tested it with run, however the APIs seems to be working fine used here.

- Have you tested all the tables with the fine-grained access token?
- If so, could you kindly attach the query results in the PR body for reference?

Thanks!

PR #464, opened 2024-12-18
Add `run_attempt` as an optional key column for `github_actions_repository_workflow_run` get calls

tsibley:
_builds upon #463, but that PR can be considered separately from this one_

This is the bare minimum to allow previous run attempts to be queried, which is useful for comparisons between attempts. User queries must explicitly iterate over a run's multiple attempts as necessary, e.g. by parsing `previous_attempt_url` in a `with recursive` CTE or simply presuming the series of attempts is `generate_series(1, run_attempt)`.

A future improvement for this plugin might be exposing a new table, e.g. `github_actions_repository_workflow_run_attempt`, with the same list key columns of `repository_full_name` and `id`, and handling the iteration internally. GitHub's API unfortunately provides no listing endpoints for run attempts.

### DIFF
--- a/github/table_github_actions_repository_workflow_run.go
+++ b/github/table_github_actions_repository_workflow_run.go
@@ -8,6 +8,7 @@ import (
 	"github.com/turbot/steampipe-plugin-sdk/v5/grpc/proto"
 	"github.com/turbot/steampipe-plugin-sdk/v5/plugin"
 	"github.com/turbot/steampipe-plugin-sdk/v5/plugin/transform"
+	"github.com/turbot/steampipe-plugin-sdk/v5/query_cache"
 )
 
 func tableGitHubActionsRepositoryWorkflowRun() *plugin.Table {
@@ -26,7 +27,11 @@ func tableGitHubActionsRepositoryWorkflowRun() *plugin.Table {
 			},
 		},
 		Get: &plugin.GetConfig{
-			KeyColumns:        plugin.AllColumns([]string{"repository_full_name", "id"}),
+			KeyColumns: []*plugin.KeyColumn{
+				{Name: "repository_full_name", Require: plugin.Required, Operators: []string{"="}},
+				{Name: "id", Require: plugin.Required, Operators: []string{"="}},
+				{Name: "run_attempt", Require: plugin.Optional, Operators: []string{"="}, CacheMatch: query_cache.CacheMatchExact},
+			},
 			ShouldIgnoreError: isNotFoundError([]string{"404"}),
 			Hydrate:           tableGitHubRepoWorkflowRunGet,
 		},
@@ -141,6 +146,7 @@ func tableGitHubRepoWorkflowRunList(ctx context.Context, d *plugin.QueryData, h 
 
 func tableGitHubRepoWorkflowRunGet(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
 	runId := d.EqualsQuals["id"].GetInt64Value()
+	runAttempt := int(d.EqualsQuals["run_attempt"].GetInt64Value())
 	orgName := d.EqualsQuals["repository_full_name"].GetStringValue()
 
 	// Empty check for the parameters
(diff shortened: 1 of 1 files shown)


misraved (approved):

github/table_github_actions_repository_workflow_run.go
```diff
@@ -26,7 +26,11 @@ func tableGitHubActionsRepositoryWorkflowRun() *plugin.Table {
 			},
 		},
 		Get: &plugin.GetConfig{
-			KeyColumns:        plugin.AllColumns([]string{"repository_full_name", "id"}),
+			KeyColumns: []*plugin.KeyColumn{
+				{Name: "repository_full_name", Require: plugin.Required, Operators: []string{"="}},
+				{Name: "id", Require: plugin.Required, Operators: []string{"="}},
+				{Name: "run_attempt", Require: plugin.Optional, Operators: []string{"="}, CacheMatch: "exact"},
```
misraved, line 32: @tsibley is there a specific reason for adding `CacheMatch: "exact"` to this optional qual? It is typically used for columns that are populated using `FromQuals`.

ParthaI: @misraved, @tsibley, 

From my understanding, we don’t need to include `CacheMatch: "exact"` for these optional qualifiers here. This is because the `run_attempt` optional qualifier is part of the `Get` config, not the `List` config.

According to the table design, the `Get` API call should always return a single row. In this table, the unique identifier is the `id` column, which is unique per row. 

Query, result:

List Config execution:
```sql

> select id, run_attempt from github_actions_repository_workflow_run where repository_full_name = 'turbot/steampipe-plugin-aws' and  run_attempt > 2
+-------------+-------------+
| id          | run_attempt |
+-------------+-------------+
| 11666032178 | 3           |
| 10667532782 | 3           |
| 9872917837  | 5           |
| 8801296800  | 3           |
| 8753110746  | 3           |
| 8296554622  | 3           |
| 8065062549  | 4           |
| 8076825093  | 3           |
| 7887624205  | 3           |
| 7358331805  | 4           |
+-------------+-------------+
```

Get Config execution: 
```sql
> select id, run_attempt from github_actions_repository_workflow_run where repository_full_name = 'turbot/steampipe-plugin-aws' and id = 7358331805 and run_attempt = 4
+------------+-------------+
| id         | run_attempt |
+------------+-------------+
| 7358331805 | 4           |
+------------+-------------+
```

tsibley: @ParthaI `id` is unique per row only because List can only return the latest `run_attempt`. A single id has multiple run attempts, though, and so the Get call needs to consider it as part of the key.

@misraved Without `CacheMatch: "exact"`, there are false positive cache hits on a Get after a List when accessing previous run attempts.

ParthaI: > @ParthaI id is unique per row only because List can only return the latest run_attempt. A single id has multiple run attempts, though, and so the Get call needs to consider it as part of the key.

Apologies for the misunderstanding, @tsibley. I was unaware of the API behaviour. It makes perfect sense to include `CacheMatch: "exact"` for the `run_attempt` optional key column qualifier. Thank you for clarifying!

Query result WITHOUT `CacheMatch: "exact"`:

```sql
> select id, run_attempt from github_actions_repository_workflow_run where repository_full_name = 'turbot/steampipe-plugin-aws' and id = 7358331805
+------------+-------------+
| id         | run_attempt |
+------------+-------------+
| 7358331805 | 4           |
+------------+-------------+

Time: 1.7s. Rows returned: 0. Rows fetched: 1. Hydrate calls: 0.
> select id, run_attempt from github_actions_repository_workflow_run where repository_full_name = 'turbot/steampipe-plugin-aws' and id = 7358331805 and run_attempt = 1
+----+-------------+
| id | run_attempt |
+----+-------------+
+----+-------------+

Time: 20ms. Rows returned: 0. Rows fetched: 1 (cached). Hydrate calls: 0.
```

Query result WITH `CacheMatch: "exact"`:

```sql

> select id, run_attempt from github_actions_repository_workflow_run where repository_full_name = 'turbot/steampipe-plugin-aws' and id = 7358331805
+------------+-------------+
| id         | run_attempt |
+------------+-------------+
| 7358331805 | 4           |
+------------+-------------+

Time: 1.7s. Rows returned: 0. Rows fetched: 1. Hydrate calls: 0.
> select id, run_attempt from github_actions_repository_workflow_run where repository_full_name = 'turbot/steampipe-plugin-aws' and id = 7358331805 and run_attempt = 1
+------------+-------------+
| id         | run_attempt |
+------------+-------------+
| 7358331805 | 1           |
+------------+-------------+

Time: 451ms. Rows returned: 0. Rows fetched: 1. Hydrate calls: 0.
```

Additionally, could you please use `CacheMatch: query_cache.CacheMatchExact` instead of `CacheMatch: "exact"`? For reference, you can check this example: [AWS CloudControl Resource Table](https://github.com/turbot/steampipe-plugin-aws/blob/7709c95b6b4b0c3e61307634f7dbda00499b91d7/aws/table_aws_cloudcontrol_resource.go#L33).

tsibley: I’ve [swapped out the string literal for the symbol](https://github.com/turbot/steampipe-plugin-github/compare/f2417d9f13830e627a02347137563abb52b66093..fc3ed5f06b7873947623d5f31595d5a327ae3ea8). Thanks for the pointer to it. I’d originally looked for it (quickly) but didn’t find it and the explanatory code comments used the string literal so I used that too.

PR #470, opened 2025-01-23
github_workflow: Check HTTP status code, not a stringy error message

tsibley:
This checked started failing because (it seems) GitHub changed their error message string from "No commit found" to "No object found".  No longer ignoring the 404 led to query errors such as:

    ERROR:  rpc error: code = Unknown desc = github: GET https://api.github.com/repos/nextstrain/ncov/contents/.github/workflows/trs-tmp.yaml?ref=master: 404 No object found for the path .github/workflows/trs-tmp.yaml []

See <https://github.com/nextstrain/status/issues/36> for the debugging that came to this conclusion.

Checking stringy error messages is doomed to this sort of problem from the start.  This, unfortunately, is just one place that's done: there are many others in this codebase (grep for "err.Error").  I'm leaving them be for now, but they suffer from the same bug waiting to happen.

# Example query results
<details>
  <summary>Results</summary>
  
```
Add example SQL query results here (please include the input queries as well)
```
</details>


ParthaI (approved):
LGTM

PR #546, opened 2026-05-20
chore: bump SDK to v6.0.0 + migrate Config access through GetConfig

kaidaguerre:
Migrates `connection.Config` direct reads to `connection.GetConfig()` per `steampipe-plugin-sdk` v6.0.0 (see [steampipe-plugin-sdk#938](https://github.com/turbot/steampipe-plugin-sdk/pull/938)). GitHub-side rollout of the same fix that landed for AWS in [steampipe-plugin-aws#2756](https://github.com/turbot/steampipe-plugin-aws/pull/2756). The github plugin had the highest-traffic uncached `connection.Config` read site of any plugin sampled (`appendUserInteractionAbilityForIssue` in `user_utils.go`, called per-query on 7+ tables) — this PR closes that race surface. Build, vet, test, and `go test -race` all green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## v1.8.2 [2026-05-20]
+
+_Bug fixes_
+
+- Fixed `ExpiredToken`-style errors on long-running queries when Turbot Pipes refreshes GitHub App installation tokens mid-query. The plugin now reads the connection config via the SDK's `GetConfig` accessor (which holds a read lock) on every API call, so in-flight goroutines pick up rotated credentials. ([#<PR>](https://github.com/turbot/steampipe-plugin-github/pull/<PR>))
+
+_Dependencies_
+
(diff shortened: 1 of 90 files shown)


kaidaguerre:
Closing — this PR does not fix a customer-visible bug.

Cross-checking the v6 SDK race fix against this plugin's auth model: Pipes does NOT push rotated credentials into the github plugin's connection config. GitHub auth uses either `token` (PAT — long-lived static) or `app_id` + `app_installation_id` + `app_private_key` (the App private key is long-lived; the go-github-app library mints and refreshes ~1-hour installation tokens locally from it). Pipes has no rotation cycle that would mutate `Connection.Config` mid-query for this plugin.

The CHANGELOG bug-fix narrative claiming `ExpiredToken`-style fixes for credential rotation was therefore misleading. The SDK bump on its own is not harmful, but shipping it under a false bug-fix narrative is.

For context, see https://github.com/turbot/steampipe-plugin-aws/pull/2756 — the actual rotation fix adds a custom `aws.CredentialsProvider` that re-reads `connection.GetConfig()` per signing request. The v6 SDK race fix is supporting infrastructure for that pattern, not a fix on its own.

If a pure dependency-hygiene bump becomes desirable later, a new PR with a `_Dependencies_`-only CHANGELOG entry would be the right shape.

PR #548, opened 2026-06-19
Add the `github_project_v2` table

dark-panda:
This PR adds a table called `github_project_v2` that contains some of the fields exposed by the GraphQL API for ProjectV2 and adds a few fields to `github_issue` to allow you to join the two tables together. We're using the `github_project_v2` name here rather than `github_project` because of existing fields in tables like `github_team` and `github_organization` that reference `projects_v2_total_count` and such. 

Example queries:

```
=> select * from github_project_v2 where organization = 'turbot' ;
-[ RECORD 1 ]------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------
login_id                 | MDQ6VXNlcjg0Nzgz
organization             | turbot
number                   | 14
id                       | 10217986
node_id                  | PVT_kwDOAlEJmM4Am-oC
owner                    | {"type": "Organization", "user": {"id": 38865304, "login": "turbot"}, "organization": {"id": 38865304, "login": "turbot"}}
creator                  | {"url": "https://github.com/apps/github-project-automation", "login": "github-project-automation", "avatar_url": "https://avatars.githubusercontent.com/in/235829?v=4"}
title                    | Issue Management
description              |
is_public                | t
closed_at                | 2026-06-08 13:57:12+00
created_at               | 2024-08-29 03:57:44+00
updated_at               | 2026-06-08 13:57:12+00
state                    | closed
latest_status_update     | [NULL]
is_template              | f
readme                   | Tracking all examples required by end users
resource_path            | /orgs/turbot/projects/14
url                      | https://github.com/orgs/turbot/projects/14
repositories             | ["turbot/guardrails-samples"]
repositories_total_count | 1
teams                    | ["", ""]
teams_total_count        | 2
sp_connection_name       | github
sp_ctx                   | {"steampipe": {"sdk_version": "5.14.0"}, "connection_name": "github"}
_ctx                     | {"steampipe": {"sdk_version": "5.14.0"}, "connection_name": "github"}

=> select
  i.id,
  i.title,
  p.title as project_title,
  p.state as project_state,
  p.id as project_id
from
  github_issue i,
  jsonb_array_elements_text(i.project_items) as project_node_id
  join github_project_v2 p on p.node_id = project_node_id
where
  i.repository_full_name = 'turbot/guardrails-samples'
  and p.organization = 'turbot';
     id     |                           title                            |  project_title   | project_state | project_id
------------+------------------------------------------------------------+------------------+---------------+------------
  815566711 | SecHub - Feature - SecHub needs a heartbeat                | Issue Management | closed        |   10217986
  841771067 | GraphQL - Get the details of a policy such as descriptions | Issue Management | closed        |   10217986
  966425300 | Calculated Policy - Log Group Deletions                    | Issue Management | closed        |   10217986
 1016873038 | TrimBot - Fix the version for dependent libraries          | Issue Management | closed        |   10217986
  966422959 | TrimBot - Error handling should be more verbose            | Issue Management | closed        |   10217986
  966420924 | TrimBot - SNS topics not all removed                       | Issue Management | closed        |   10217986
(6 rows)
```

### DIFF
--- a/docs/tables/github_project_v2.md
+++ b/docs/tables/github_project_v2.md
@@ -1,0 +1,209 @@
+---
+title: "Steampipe Table: github_project_v2 - Query GitHub Projects (V2) using SQL"
+description: "Allows users to query GitHub Projects (V2), providing insights into the organization-level projects used to plan and track work."
+folder: "Project"
+---
+
+# Table: github_project_v2 - Query GitHub Projects (V2) using SQL
+
+GitHub Projects (V2) is GitHub's flexible, table and board based tool for planning and tracking work across issues and pull requests. It allows teams to organize work items, track status, and view progress across one or more repositories using customizable views, fields, and workflows.
+
+## Table Usage Guide
+
+The `github_project_v2` table provides insights into the ProjectsV2 owned by a GitHub organization. As a project manager or developer, explore project-specific details through this table, including title, description, visibility, status, linked repositories, and linked teams. Utilize it to uncover information about projects, such as which ones are public, which repositories and teams are associated with them, and when they were last updated.
+
+To query this table using a [fine-grained access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#creating-a-fine-grained-personal-access-token), the following permissions are required:
+  - Organization permissions:
+    - Projects (Read-only): Required to access all columns.
+
+**Important Notes**
+- You must specify the `organization` column in a `where` or `join` clause to query the table.
+
+## Examples
+
+### List the projects in an organization
+Explore the title, state, and visibility of the ProjectsV2 owned by a specific GitHub organization to get an overview of ongoing work.
+
+```sql+postgres
+select
+  organization,
+  number,
+  title,
+  state,
+  is_public,
+  created_at
+from
+  github_project_v2
+where
+  organization = 'turbot';
+```
+
+```sql+sqlite
+select
+  organization,
+  number,
+  title,
+  state,
+  is_public,
+  created_at
+from
+  github_project_v2
+where
+  organization = 'turbot';
+```
+
+### List open projects in an organization
+Identify the projects that are still open in a specific organization, to help focus attention on active planning boards.
+
+```sql+postgres
+select
+  organization,
+  number,
+  title,
+  created_at,
+  updated_at
+from
+  github_project_v2
+where
+  organization = 'turbot'
+and
+  state = 'open';
+```
+
+```sql+sqlite
+select
+  organization,
+  number,
+  title,
+  created_at,
+  updated_at
+from
+  github_project_v2
+where
+  organization = 'turbot'
+  and state = 'open';
+```
+
+### Get a specific project by number
+Retrieve the details of a single project using its project number, useful when you already know which project you want to inspect.
+
+```sql+postgres
+select
+  number,
+  title,
+  description,
+  owner,
+  creator
+from
+  github_project_v2
+where
+  organization = 'turbot'
+and
+  number = 1;
+```
+
+```sql+sqlite
+select
+  number,
+  title,
+  description,
+  owner,
+  creator
+from
+  github_project_v2
+where
+  organization = 'turbot'
+  and number = 1;
+```
+
+### List projects updated in the last 30 days
+Discover the projects that have had recent activity, useful for tracking which planning boards are actively being maintained.
+
+```sql+postgres
+select
+  number,
+  title,
+  updated_at
+from
+  github_project_v2
+where
+  organization = 'turbot'
+and
+  updated_at >= now() - interval '30 days'
+order by
+  updated_at desc;
+```
+
+```sql+sqlite
+select
+  number,
+  title,
+  updated_at
+from
+  github_project_v2
+where
+  organization = 'turbot'
+  and updated_at >= datetime('now', '-30 days')
+order by
+  updated_at desc;
+```
+
+### List repositories and teams linked to each project
+Explore which repositories and teams are linked to each project, to understand the scope of collaboration around a project.
+
+```sql+postgres
+select
+  number,
+  title,
+  repositories,
+  repositories_total_count,
+  teams,
+  teams_total_count
+from
+  github_project_v2
+where
+  organization = 'turbot';
+```
+
+```sql+sqlite
+select
+  number,
+  title,
+  repositories,
+  repositories_total_count,
+  teams,
+  teams_total_count
+from
+  github_project_v2
+where
+  organization = 'turbot';
+```
+
+### List the latest status update for each project
+Explore the most recent status update posted on each project, useful for quickly checking the reported health and progress of a project.
+
+```sql+postgres
+select
+  number,
+  title,
+  latest_status_update ->> 'status' as status,
+  latest_status_update ->> 'body' as body,
+  latest_status_update ->> 'created_at' as reported_at
+from
+  github_project_v2
+where
+  organization = 'turbot';
+```
+
+```sql+sqlite
+select
+  number,
+  title,
+  json_extract(latest_status_update, '$.status') as status,
+  json_extract(latest_status_update, '$.body') as body,
+  json_extract(latest_status_update, '$.created_at') as reported_at
+from
+  github_project_v2
+where
+  organization = 'turbot';
+```
--- a/github/models/misc.go
+++ b/github/models/misc.go
@@ -1,6 +1,7 @@
 package models
 
 import (
+	"strings"
 	"time"
 
 	"github.com/shurcooL/githubv4"
@@ -16,6 +17,35 @@ func (t NullableTime) MarshalJSON() ([]byte, error) {
 	} else {
 		return t.Time.MarshalJSON()
 	}
+}
+
+// NullableDate wraps time.Time to support unmarshalling the GraphQL `Date`
+// scalar (e.g. "2024-08-11"), which does not include a time-of-day or
+// timezone component like `DateTime`/`ISO8601DateTime` scalars do.
+type NullableDate struct {
+	time.Time
+}
+
+func (d *NullableDate) UnmarshalJSON(data []byte) error {
+	s := strings.Trim(string(data), `"`)
+	if s == "" || s == "null" {
+		d.Time = time.Time{}
+		return nil
+	}
+
+	parsed, err := time.Parse("2006-01-02", s)
+	if err != nil {
+		return err
+	}
+	d.Time = parsed
+	return nil
+}
+
+func (d NullableDate) MarshalJSON() ([]byte, error) {
+	if d.IsZero() {
+		return []byte("null"), nil
+	}
+	return []byte(`"` + d.Time.Format("2006-01-02") + `"`), nil
 }
 
 type NameSlug struct {
@@ -55,7 +85,7 @@ type SponsorsListing struct {
 	FullDescription            string               `json:"full_description"`
 	IsPublic                   bool                 `json:"is_public"`
 	Name                       string               `json:"name"`
-	NextPayoutDate             time.Time            `json:"next_payout_date"`
+	NextPayoutDate             NullableDate         `json:"next_payout_date"`
 	ResidenceCountryOrRegion   string               `json:"residence_country_or_region"`
 	ShortDescription           string               `json:"short_description"`
 	Slug                       string               `json:"slug"`
--- a/github/models/project_v2.go
+++ b/github/models/project_v2.go
@@ -1,0 +1,61 @@
+package models
+
+// ProjectV2Owner represents the owner of a project, which can be an Organization or User.
+type ProjectV2Owner struct {
+	TypeName     string `graphql:"type: __typename" json:"type"`
+	Organization struct {
+		Id    int    `graphql:"id: databaseId" json:"id"`
+		Login string `json:"login"`
+	} `graphql:"... on Organization" json:"organization,omitempty"`
+	User struct {
+		Id    int    `graphql:"id: databaseId" json:"id"`
+		Login string `json:"login"`
+	} `graphql:"... on User" json:"user,omitempty"`
+}
+
+// ProjectV2StatusUpdate represents a single status update on a project.
+type ProjectV2StatusUpdate struct {
+	Id         string       `graphql:"id: fullDatabaseId" json:"id"`
+	NodeId     string       `graphql:"nodeId: id" json:"node_id"`
+	Status     string       `json:"status"`
+	Body       string       `json:"body"`
+	StartDate  string       `json:"start_date"`
+	TargetDate string       `json:"target_date"`
+	CreatedAt  NullableTime `json:"created_at"`
+	UpdatedAt  NullableTime `json:"updated_at"`
+	Creator    Actor        `json:"creator"`
+}
+
+type ProjectV2 struct {
+	Id                 string         `graphql:"id: fullDatabaseId @include(if:$includeId)" json:"id"`
+	NodeId             string         `graphql:"nodeId: id @include(if:$includeNodeId)" json:"node_id"`
+	Number             int            `json:"number"`
+	Owner              ProjectV2Owner `graphql:"owner @include(if:$includeOwner)" json:"owner,omitempty"`
+	Creator            Actor          `graphql:"creator @include(if:$includeCreator)" json:"creator,omitempty"`
+	Title              string         `graphql:"title @include(if:$includeTitle)" json:"title"`
+	Description        string         `graphql:"description: shortDescription @include(if:$includeDescription)" json:"description"`
+	IsPublic           bool           `graphql:"public @include(if:$includeIsPublic)" json:"public"`
+	ClosedAt           NullableTime   `graphql:"closedAt @include(if:$includeClosedAt)" json:"closed_at"`
+	CreatedAt          NullableTime   `graphql:"createdAt @include(if:$includeCreatedAt)" json:"created_at"`
+	UpdatedAt          NullableTime   `graphql:"updatedAt @include(if:$includeUpdatedAt)" json:"updated_at"`
+	Closed             bool           `graphql:"closed @include(if:$includeState)" json:"closed"`
+	LatestStatusUpdate struct {
+		Nodes []ProjectV2StatusUpdate
+	} `graphql:"statusUpdates(last: 1) @include(if:$includeLatestStatusUpdate)" json:"latest_status_update"`
+	IsTemplate   bool   `graphql:"template @include(if:$includeIsTemplate)" json:"template"`
+	Readme       string `graphql:"readme @include(if:$includeReadme)" json:"readme"`
+	ResourcePath string `graphql:"resourcePath @include(if:$includeResourcePath)" json:"resource_path"`
+	Url          string `graphql:"url @include(if:$includeUrl)" json:"url"`
+	Repositories struct {
+		TotalCount int
+		Nodes      []struct {
+			NameWithOwner string `json:"name_with_owner"`
+		}
+	} `graphql:"repositories(first: 100) @include(if:$includeRepositories)" json:"repositories"`
+	Teams struct {
+		TotalCount int
+		Nodes      []struct {
+			Slug string `json:"slug"`
+		}
+	} `graphql:"teams(first: 100) @include(if:$includeTeams)" json:"teams"`
+}
--- a/github/table_github_issue.go
+++ b/github/table_github_issue.go
@@ -64,6 +64,9 @@ func sharedIssueColumns() []*plugin.Column {
 		{Name: "user_did_author", Type: proto.ColumnType_BOOL, Hydrate: issueHydrateUserDidAuthor, Transform: transform.FromValue(), Description: "If true, user authored the issue."},
 		{Name: "user_subscription", Type: proto.ColumnType_STRING, Hydrate: issueHydrateUserSubscription, Transform: transform.FromValue(), Description: "Subscription state of the user to the issue."},
 		{Name: "assignees", Type: proto.ColumnType_JSON, Hydrate: issueHydrateAssignees, Transform: transform.FromValue().NullIfZero(), Description: "A list of Users assigned to the issue."},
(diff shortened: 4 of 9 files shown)


github/table_github_project_v2.go
```diff
@@ -0,0 +1,149 @@
...
+		List: &plugin.ListConfig{
+			KeyColumns: []*plugin.KeyColumn{
+				{
+					Name:    "organization",
+					Require: plugin.Required,
+				},
+				{
+					Name:      "updated_at",
+					Require:   plugin.Optional,
+					Operators: []string{">", ">="},
```
cbruno10, line 61: Does the List function use the `updated_at` qual? Looking at the GraphQL `projectsV2` field, I don't see a way to filter on the updated at time either, so not sure if we can filter on this field at the API layer.

github/table_github_project_v2.go
```diff
@@ -0,0 +1,149 @@
...
+	tableCols := []*plugin.Column{
+		{Name: "organization", Type: proto.ColumnType_STRING, Transform: transform.FromQual("organization"), Description: "The organization name."},
+	}
+
+	return append(tableCols, sharedProjectV2Columns()...)
+}
+
+func sharedProjectV2Columns() []*plugin.Column {
+	return []*plugin.Column{
+		{Name: "number", Type: proto.ColumnType_INT, Transform: transform.FromField("Number", "Node.Number"), Description: "The project number."},
```
cbruno10, line 24: Does the API ever return the data under `Node.Number`?

github/table_github_project_v2.go
```diff
@@ -0,0 +1,149 @@
...
+				{
+					Name:      "updated_at",
+					Require:   plugin.Optional,
+					Operators: []string{">", ">="},
+				},
+			},
+			ShouldIgnoreError: isNotFoundError([]string{"404"}),
+			Hydrate:           tableGitHubProjectV2List,
+		},
+		Get: &plugin.GetConfig{
```
cbruno10, line 67: Does the `GetConfig` work correctly when passing in `organization` and `number`?

`KeyColumns` requires `organization` and `number`, but the hydrate reads `quals["id"]` (never populated, so it's 0) and `queries projectV2(id: $projectId)`. Looking at the GraphQL schema, `Organization.projectV2` only accepts `number`, so this errors server-side with `Unknown argument 'id'`. 

I think the Get function should use `quals["number"]` and use `graphql:"projectV2(number: $number)"` instead.

github/table_github_issue.go
```diff
@@ -64,6 +64,7 @@ func sharedIssueColumns() []*plugin.Column {
 		{Name: "user_did_author", Type: proto.ColumnType_BOOL, Hydrate: issueHydrateUserDidAuthor, Transform: transform.FromValue(), Description: "If true, user authored the issue."},
 		{Name: "user_subscription", Type: proto.ColumnType_STRING, Hydrate: issueHydrateUserSubscription, Transform: transform.FromValue(), Description: "Subscription state of the user to the issue."},
 		{Name: "assignees", Type: proto.ColumnType_JSON, Hydrate: issueHydrateAssignees, Transform: transform.FromValue().NullIfZero(), Description: "A list of Users assigned to the issue."},
+		{Name: "project_items", Type: proto.ColumnType_JSON, Hydrate: issueHydrateProjectItems, Transform: transform.FromValue().NullIfZero(), Description: "A list of project node IDs (PVT_...) for ProjectV2 projects the issue belongs to."},
```
cbruno10:
```suggestion
		{Name: "projects_v2", Type: proto.ColumnType_JSON, Hydrate: issueHydrateProjectItems, Transform: transform.FromValue().NullIfZero(), Description: "A list of ProjectV2 projects associated with the issue."},
```

For this column, similar to `assignees` and `labels`, I think we should return more project data, including its scalar fields as part of the project object. So we can reuse the `ProjectV2` model you already defined and include its scalar fields:
  - `id`
  - `node_id`
  - `number`
  - `title`
  - `description`
  - `state`
  - `is_public`
  - `created_at`
  - `updated_at`
  - `closed_at`

Notably, we should skip the pageable connections (items/fields/views), and `@include`-gate anything heavy like `latest_status_update`. Users can still join on `node_id`, but now get titles/state without one.

Can you please also add a `projects_v2_total_count` column with the description `Count of ProjectV2 projects associated with the issue.`? The model already fetches `ProjectItems.TotalCount`, so it's basically free, and it matches the existing `projects_v2_total_count` columns in the `github_organization` and `github_team` tables.

cbruno10 (changes requested):
@dark-panda Thanks for raising this PR! I've added a few suggestions and questions in this PR. A few other minor points:
- Can you please add a table doc page for `docs/tables/github_project_v2.md`?
- Can you please also update the PR description, which have different (possibly older) columns? I think it'd be helpful to see the results of the example queries from the upcoming table doc.
- Locally, `models/project_v2.go` and `table_github_project_v2.go` both fail `gofmt -l`, can you please check these files?
- 
If you have any questions, please let us know!

github/table_github_project_v2.go
dark-panda: Yes, it's returning an integer for me.

github/table_github_project_v2.go
dark-panda: Right, I've updated this -- it returns as one would expect now:

```
steampipe:steampipe@127=> select * from github.github_project_v2 where organization = 'turbot' and number = 14;
-[ RECORD 1 ]------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------
login_id                 | MDQ6VXNlcjg0Nzgz
organization             | turbot
number                   | 14
id                       | 10217986
node_id                  | PVT_kwDOAlEJmM4Am-oC
owner                    | {"type": "Organization", "user": {"id": 38865304, "login": "turbot"}, "organization": {"id": 38865304, "login": "turbot"}}
creator                  | {"url": "https://github.com/apps/github-project-automation", "login": "github-project-automation", "avatar_url": "https://avatars.githubusercontent.com/in/235829?v=4"}
title                    | Issue Management
description              |
is_public                | t
closed_at                | 2026-06-08 13:57:12+00
created_at               | 2024-08-29 03:57:44+00
updated_at               | 2026-06-08 13:57:12+00
state                    | closed
latest_status_update     | !NULL!
is_template              | f
readme                   | Tracking all examples required by end users
resource_path            | /orgs/turbot/projects/14
url                      | https://github.com/orgs/turbot/projects/14
repositories             | ["turbot/guardrails-samples"]
repositories_total_count | 1
teams                    | ["", ""]
teams_total_count        | 2
sp_connection_name       | github
sp_ctx                   | {"steampipe": {"sdk_version": "5.14.0"}, "connection_name": "github"}
_ctx                     | {"steampipe": {"sdk_version": "5.14.0"}, "connection_name": "github"}

Time: 10.615 ms

steampipe:steampipe@127=> select * from github.github_project_v2 where organization = 'turbot' and number = 1400;
(0 rows)

Time: 225.018 ms
```

github/table_github_project_v2.go
dark-panda: This is a client-side-early-bail-out, so while we can't filter on `updated_at`, we can ask the API to return in order of `updated_at` and then return early when we've got the results we need.

github/models/project_v2.go
```diff
@@ -0,0 +1,61 @@
...
+	Title              string         `graphql:"title @include(if:$includeTitle)" json:"title"`
+	Description        string         `graphql:"description: shortDescription @include(if:$includeDescription)" json:"description"`
+	IsPublic           bool           `graphql:"public @include(if:$includeIsPublic)" json:"public"`
+	ClosedAt           NullableTime   `graphql:"closedAt @include(if:$includeClosedAt)" json:"closed_at"`
+	CreatedAt          NullableTime   `graphql:"createdAt @include(if:$includeCreatedAt)" json:"created_at"`
+	UpdatedAt          NullableTime   `graphql:"updatedAt @include(if:$includeUpdatedAt)" json:"updated_at"`
+	Closed             bool           `graphql:"closed @include(if:$includeState)" json:"closed"`
+	LatestStatusUpdate struct {
+		Nodes []ProjectV2StatusUpdate
+	} `graphql:"statusUpdates(last: 1) @include(if:$includeLatestStatusUpdate)" json:"latest_status_update"`
```
cbruno10: This returns the oldest status update, not the latest. The `statusUpdates` connection defaults to `orderBy: {field: CREATED_AT, direction: DESC}`, so `last: 1` picks the earliest one. The "List the latest status update" doc example is wrong today for any project with more than one update.

```suggestion
	} `graphql:"statusUpdates(first: 1) @include(if:$includeLatestStatusUpdate)" json:"latest_status_update"`
```

github/table_github_issue.go
```diff
@@ -64,6 +64,9 @@ func sharedIssueColumns() []*plugin.Column {
...
 		{Name: "user_subscription", Type: proto.ColumnType_STRING, Hydrate: issueHydrateUserSubscription, Transform: transform.FromValue(), Description: "Subscription state of the user to the issue."},
 		{Name: "assignees", Type: proto.ColumnType_JSON, Hydrate: issueHydrateAssignees, Transform: transform.FromValue().NullIfZero(), Description: "A list of Users assigned to the issue."},
+		{Name: "project_items", Type: proto.ColumnType_JSON, Hydrate: issueHydrateProjectItems, Transform: transform.FromValue(), Description: "A list of project node IDs (PVT_...) for ProjectV2 projects the issue belongs to."},
+		{Name: "project_items_total_count", Type: proto.ColumnType_INT, Hydrate: issueHydrateProjectItemsTotalCount, Transform: transform.FromValue(), Description: "Count of ProjectV2 projects the issue belongs to."},
+		{Name: "projects_v2_total_count", Type: proto.ColumnType_INT, Hydrate: issueHydrateProjectsV2TotalCount, Transform: transform.FromValue(), Description: "Count of ProjectsV2 the issue is linked to."},
```
cbruno10: This is only part of what I asked for in the earlier review. `project_items` is a bare list of `PVT_...` node IDs, which is hard to use without another lookup, and `project_items_total_count` / `projects_v2_total_count` are near duplicates (`projectItems` defaults to `includeArchived: true` and counts items, `projectsV2` counts projects).

Could you replace these with a `projects_v2` JSON column carrying the scalar project fields (id, node_id, number, title, closed, url, etc.) plus a single `projects_v2_total_count`?

github/models/project_v2.go
```diff
@@ -0,0 +1,61 @@
...
+	TypeName     string `graphql:"type: __typename" json:"type"`
+	Organization struct {
+		Id    int    `graphql:"id: databaseId" json:"id"`
+		Login string `json:"login"`
+	} `graphql:"... on Organization" json:"organization,omitempty"`
+	User struct {
+		Id    int    `graphql:"id: databaseId" json:"id"`
+		Login string `json:"login"`
+	} `graphql:"... on User" json:"user,omitempty"`
+}
```
cbruno10: shurcooL/graphql populates every inline fragment struct whose field names match, so an org-owned project renders as `{"type":"Organization","user":{...},"organization":{...}}` with the same data in both (visible in your example output). `omitempty` does not help with structs.

Suggest flattening to `{type, id, login}`, either by selecting `databaseId`/`login` once in a flat struct or via a small hydrate.

github/table_github_project_v2.go
```diff
@@ -0,0 +1,209 @@
...
+	tableCols := []*plugin.Column{
+		{Name: "organization", Type: proto.ColumnType_STRING, Transform: transform.FromQual("organization"), Description: "The organization name."},
+	}
+
+	return append(tableCols, sharedProjectV2Columns()...)
+}
+
+func sharedProjectV2Columns() []*plugin.Column {
+	return []*plugin.Column{
+		{Name: "number", Type: proto.ColumnType_INT, Transform: transform.FromField("Number", "Node.Number"), Description: "The project number."},
```
cbruno10: Items are streamed as `models.ProjectV2` directly, never wrapped in an edge, so the `Node.Number` fallback is dead.

```suggestion
		{Name: "number", Type: proto.ColumnType_INT, Transform: transform.FromField("Number"), Description: "The project number."},
```

github/table_github_project_v2.go
```diff
@@ -0,0 +1,209 @@
...
+func tableGitHubProjectV2() *plugin.Table {
+	return &plugin.Table{
+		Name:        "github_project_v2",
+		Description: "GitHub Projects are used to organize and manage work on GitHub.",
+		List: &plugin.ListConfig{
+			KeyColumns: []*plugin.KeyColumn{
+				{
+					Name:    "organization",
+					Require: plugin.Required,
+				},
```
cbruno10: The table is org-only right now, but `User.projectsV2` and `Repository.projectsV2` also exist. If user-owned projects are ever wanted, `organization` as the required qual will not extend cleanly.

If org-only is the intent, that is fine, but please say so in the table `Description` so users know why their personal projects do not show up.

github/table_github_project_v2.go
```diff
@@ -0,0 +1,209 @@
...
+	plugin.Logger(ctx).Debug(rateLimitLogString("github_project_v2", &query.RateLimit))
+	if err != nil {
+		plugin.Logger(ctx).Error("github_project_v2", "api_error", err)
+		// GitHub's GraphQL API returns a "Could not resolve to a ProjectV2 ..."
+		// error (rather than a 404) when the project number doesn't exist. Treat
+		// this the same as a not-found result rather than propagating the error,
+		// since letting a Get error out (rather than a clean "no rows") can leave
+		// the plugin SDK's query cache with an orphaned pending entry that causes
+		// subsequent identical queries to hang waiting on it.
+		if strings.Contains(err.Error(), "Could not resolve to a ProjectV2") {
```
cbruno10: A bad org login returns `Could not resolve to an Organization with the login of '...'` (NOT_FOUND) from both List and Get, and neither path swallows it, so users get a raw error instead of zero rows. `github_organization` handles this case; worth matching it here.

github/models/project_v2.go
```diff
@@ -0,0 +1,61 @@
...
+	IsTemplate   bool   `graphql:"template @include(if:$includeIsTemplate)" json:"template"`
+	Readme       string `graphql:"readme @include(if:$includeReadme)" json:"readme"`
+	ResourcePath string `graphql:"resourcePath @include(if:$includeResourcePath)" json:"resource_path"`
+	Url          string `graphql:"url @include(if:$includeUrl)" json:"url"`
+	Repositories struct {
+		TotalCount int
+		Nodes      []struct {
+			NameWithOwner string `json:"name_with_owner"`
+		}
+	} `graphql:"repositories(first: 100) @include(if:$includeRepositories)" json:"repositories"`
```
cbruno10: `repositories` and `teams` (line 60) are capped at 100 with no pagination while the `*_total_count` columns report the true count. That is acceptable, but the column descriptions should say "first 100" so the mismatch is not a surprise.

github/table_github_project_v2.go
```diff
@@ -0,0 +1,209 @@
...
+			}
+		}
+	}
+
+	var query struct {
+		RateLimit    models.RateLimit
+		Organization struct {
+			ProjectsV2 struct {
+				PageInfo   models.PageInfo
+				TotalCount int
```
cbruno10: Fetched but never read. Can be dropped.

docs/tables/github_project_v2.md
```diff
@@ -0,0 +1,209 @@
...
+
+GitHub Projects (V2) is GitHub's flexible, table and board based tool for planning and tracking work across issues and pull requests. It allows teams to organize work items, track status, and view progress across one or more repositories using customizable views, fields, and workflows.
+
+## Table Usage Guide
+
+The `github_project_v2` table provides insights into the ProjectsV2 owned by a GitHub organization. As a project manager or developer, explore project-specific details through this table, including title, description, visibility, status, linked repositories, and linked teams. Utilize it to uncover information about projects, such as which ones are public, which repositories and teams are associated with them, and when they were last updated.
+
+To query this table using a [fine-grained access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#creating-a-fine-grained-personal-access-token), the following permissions are required:
+  - Organization permissions:
+    - Projects (Read-only): Required to access all columns.
```
cbruno10: Your own output shows `teams` coming back as `["", ""]`. `Team.slug` is `String!`, so those are not null slugs, most likely the token lacked org Members read and GitHub returned redacted nodes. That contradicts "Required to access all columns" here.

Could you re-test `teams` with a token that has org Members read and update this section? Classic PATs also need the `read:project` scope, which should be listed too.

docs/tables/github_project_v2.md
```diff
@@ -0,0 +1,209 @@
...
+```sql+postgres
+select
+  organization,
+  number,
+  title,
+  created_at,
+  updated_at
+from
+  github_project_v2
+where
```
cbruno10: Nit: the postgres blocks use `where\n  x\nand\n  y` while the sqlite twins use the repo's `  and y` style. Same at lines 99 and 130.

github/models/misc.go
```diff
@@ -55,7 +85,7 @@ type SponsorsListing struct {
 	FullDescription            string               `json:"full_description"`
 	IsPublic                   bool                 `json:"is_public"`
 	Name                       string               `json:"name"`
-	NextPayoutDate             time.Time            `json:"next_payout_date"`
+	NextPayoutDate             NullableDate         `json:"next_payout_date"`
```
cbruno10: This is a legitimate fix (the schema field is the `Date` scalar, which `time.Time` cannot unmarshal, so `sponsors_listing` errored whenever a payout date was set), but it is unrelated to the project table. Please move it to its own PR so it can get a changelog bug-fix line. Note the JSON value also changes from RFC3339 to `"2024-08-11"` and from `"0001-01-01T00:00:00Z"` to `null` when unset.

github/models/project_v2.go
cbruno10, line 44:
```suggestion
	} `graphql:"statusUpdates(first: 1) @include(if:$includeLatestStatusUpdate)" json:"latest_status_update"`
```

The connection defaults to `CREATED_AT DESC`, so `last: 1` gives the oldest update rather than the latest one. The "latest status update" example in the docs would return the wrong row today for any project with more than one update.

github/table_github_issue.go
cbruno10, line 69: This gets partway to what I was after in the last review. A list of `PVT_...` node IDs isn't very usable on its own, and `project_items_total_count` and `projects_v2_total_count` end up being almost the same number (one counts items including archived, the other counts projects).

Could we have a `projects_v2` JSON column with the scalar project fields (id, number, title, closed, url and so on) plus a single `projects_v2_total_count` instead?

github/models/project_v2.go
cbruno10, line 14: shurcooL fills in every fragment whose field names match, so `owner` comes back with the same data twice. Running this against `turbot` gives:

```json
{"organization":{"id":38865304,"login":"turbot"},"type":"Organization","user":{"id":38865304,"login":"turbot"}}
```

Can we flatten this to `{type, id, login}`? Selecting `databaseId`/`login` once in a flat struct would do it.

github/table_github_project_v2.go
cbruno10, line 26:
```suggestion
		{Name: "number", Type: proto.ColumnType_INT, Transform: transform.FromField("Number"), Description: "The project number."},
```

We stream `models.ProjectV2` directly, never wrapped in an edge, so I don't believe the `Node.Number` fallback fires. Have you seen a need for this though?

github/table_github_project_v2.go
cbruno10, line 59: Is org-only intentional? `User.projectsV2` and `Repository.projectsV2` exist too, and with `organization` as the required qual we can't extend to those later without a breaking change. If org-only is the plan for now, that's fine, but let's say so in the table description so people know why their personal projects don't show up.

github/table_github_project_v2.go
cbruno10, line 202: A bad org login comes back as a raw error rather than zero rows:

```
select number, title from github_project_v2 where organization = 'org-that-does-not-exist'
Error: Could not resolve to an Organization with the login of 'org-that-does-not-exist'.
```

The `github_organization` table doesn't pass that error on, could we have the same behaviour here (return zero rows instead of the error)?

github/models/project_v2.go
cbruno10, line 54: `repositories` and `teams` (below) stop at 100 while the `*_total_count` columns report the real number. That's fine, but can we say "first 100" in the column descriptions so the mismatch isn't a surprise?

github/table_github_project_v2.go
cbruno10, line 113: Is `TotalCount` used, do we need it?

docs/tables/github_project_v2.md
cbruno10, line 17: The `["", ""]` in your example output is GitHub redacting team nodes the token can't see rather than empty slugs (`Team.slug` is `String!`). With an org admin token against `turbot` I get real slugs, e.g. `["all-turbot"]` on project 25.

Can we note here that `teams` needs org Members read?

docs/tables/github_project_v2.md
cbruno10: Nit: the postgres blocks use `where\n  x\nand\n  y` while the sqlite twins use the repo's `  and y` style. Same at lines 99 and 130.

github/models/misc.go
cbruno10, line 88: Good catch, `Date` can't unmarshal into `time.Time` so `sponsors_listing` errors whenever a payout date is set. It's unrelated to this table though, so could you move it to its own PR? That way it gets its own changelog entry. Worth noting there that the JSON changes from RFC3339 to `"2024-08-11"` and from `"0001-01-01T00:00:00Z"` to `null` when unset.

cbruno10 (changes requested):
@dark-panda Thanks for the latest updates and sorry for the late response. I've left a few additional review comments, can you please have a look when you get a chance? Thanks!

dark-panda:
@cbruno10 feedback applied, let's see how this looks now.

PR #552, opened 2026-06-30
Add github_organization_credential_authorization table

Kanthi-Hegde:
**Description**

The plugin currently has no way to query classic personal access tokens (PATs) that members have authorized against an organization. This PR adds a new `github_organization_credential_authorization` table to close that gap.

It wraps GitHub's [List SAML SSO authorizations for an organization](https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/orgs#list-saml-sso-authorizations-for-an-organization) REST API (`GET /orgs/{org}/credential-authorizations`), which is the only API GitHub exposes for classic PATs at the org level. The same endpoint also returns authorized SSH keys, distinguishable via the `credential_type` column.

**What's included**

- New table: `github_organization_credential_authorization`
- Registration in `plugin.go`
- Docs page with usage guide, required permissions, and example queries

**Columns**

`login`, `credential_id`, `token_last_eight`, `scopes`, `credential_type`, `credential_authorized_at`, `credential_accessed_at`, `authorized_credential_id`, `authorized_credential_note`, `authorized_credential_expires_at`, `authorized_credential_title`, `fingerprint`.

**Notes**

- The `organization` qualifier is required.
- This endpoint requires an organization owner token (`admin:org`) and is only available for organizations on GitHub Enterprise Cloud with SAML SSO enabled; other orgs return no rows. A `403`/`404` is treated as an empty result, consistent with other org-scoped tables in this plugin.


github/table_github_organization_credential_authorization.go
```diff
@@ -0,0 +1,87 @@
...
+
+func tableGitHubOrganizationCredentialAuthorization() *plugin.Table {
+	return &plugin.Table{
+		Name:        "github_organization_credential_authorization",
+		Description: "Classic personal access tokens (and SSH keys) that members have authorized to access an organization's resources through SAML single sign-on (SSO).",
+		List: &plugin.ListConfig{
+			KeyColumns: []*plugin.KeyColumn{
+				{Name: "organization", Require: plugin.Required},
+			},
+			ShouldIgnoreError: isNotFoundError([]string{"404", "403"}),
```
cbruno10, line 21: Is there a reason we are ignoring 403 errors? I think ignoring these by default for the table is usually not desired behaviour, as the user wouldn't be aware of missing permissions.

github/table_github_organization_credential_authorization.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+func tableGitHubOrganizationCredentialAuthorizationList(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
+	client := connect(ctx, d)
+
+	org := d.EqualsQuals["organization"].GetStringValue()
+
+	// Empty check
+	if org == "" {
```
cbruno10, line 48: Do we ever hit this if `organization` is a required qual for queries?

docs/tables/github_organization_credential_authorization.md
```diff
@@ -0,0 +1,209 @@
...
+
+GitHub organizations that enforce SAML single sign-on (SSO) require members to authorize their classic personal access tokens (PATs) and SSH keys before those credentials can access organization resources. A credential authorization records the token (or key) that a member has approved for use against the organization, including who it belongs to, its scopes, when it was authorized, when it was last used, and when it expires.
+
+## Table Usage Guide
+
+The `github_organization_credential_authorization` table provides insights into the classic personal access tokens authorized against a GitHub organization. As a security or compliance engineer, use this table to inventory the classic PATs that can reach your organization's resources, review the scopes they have been granted, and identify tokens that are stale, never used, or approaching expiration. The same endpoint also returns authorized SSH keys, which can be distinguished using the `credential_type` column.
+
+To query this table, the credential must be created under an organization owner. Classic personal access tokens require the `admin:org` scope (specifically `read:org`).
```
cbruno10, line 15: Having `admin:org` and `read:org` can't both be right, and GitHub's docs say `admin:org` for this endpoint.

Could you also mention it's classic PAT only (fine-grained tokens don't support it) and add the permissions block the other org tables have, e.g. github_organization_external_identity.md?

cbruno10 (changes requested):
@Kanthi-Hegde Thanks for raising this PR! I've left a few minor suggestions.

Also, can you please share the results of the example queries (just for postgres is fine) in a PR comment so we can see the shape of real results? Thanks!

github/table_github_organization_credential_authorization.go
```diff
@@ -0,0 +1,87 @@
+package github
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/go-github/v55/github"
+	"github.com/turbot/steampipe-plugin-sdk/v5/grpc/proto"
+	"github.com/turbot/steampipe-plugin-sdk/v5/plugin"
+	"github.com/turbot/steampipe-plugin-sdk/v5/plugin/transform"
```
cbruno10, line 10:
```suggestion
	"github.com/turbot/steampipe-plugin-sdk/v6/grpc/proto"
	"github.com/turbot/steampipe-plugin-sdk/v6/plugin"
	"github.com/turbot/steampipe-plugin-sdk/v6/plugin/transform"
```

`main` moved to SDK v6 in July, so can you please make this update so the plugin will build correctly after pulling in the latest from `main`?

github/table_github_organization_credential_authorization.go
cbruno10: Still think `403` should come out, as in the July review. For a security inventory table, silently returning zero rows when the token lacks `admin:org` is the worst failure mode. Also `isNotFoundError` is a substring match on the error text, so `"403"` will match any 403 body, including the SAML enforcement one.

```suggestion
			ShouldIgnoreError: isNotFoundError([]string{"404"}),
```

github/table_github_organization_credential_authorization.go
```diff
@@ -0,0 +1,87 @@
...
+
+func tableGitHubOrganizationCredentialAuthorizationList(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
+	client := connect(ctx, d)
+
+	org := d.EqualsQuals["organization"].GetStringValue()
+
+	// Empty check
+	if org == "" {
+		return nil, fmt.Errorf("'organization' qual is required for the github_organization_credential_authorization table")
+	}
```
cbruno10, line 50:
```suggestion

```

See comment above please

github/table_github_organization_credential_authorization.go
```diff
@@ -0,0 +1,87 @@
...
+
+	opts := &github.ListOptions{PerPage: 100}
```
cbruno10: In go-github v89 (#560) this becomes `&github.CredentialAuthorizationsListOptions{ListOptions: github.ListOptions{PerPage: 100}}`. That struct also adds a `Login` filter, which would make a nice optional `login` key column instead of filtering client-side.

github/table_github_organization_credential_authorization.go
```diff
@@ -0,0 +1,87 @@
...
+	if limit != nil {
+		if *limit < int64(opts.PerPage) {
+			opts.PerPage = int(*limit)
+		}
+	}
+
+	for {
+		creds, resp, err := client.Organizations.ListCredentialAuthorizations(ctx, org, opts)
+		if err != nil {
+			plugin.Logger(ctx).Error("github_organization_credential_authorization", "api_error", err)
```
cbruno10, line 64:
```suggestion
plugin.Logger(ctx).Error("github_organization_credential_authorization.tableGitHubOrganizationCredentialAuthorizationList", "api_error", err)
```

Minor one, we usually use `table.function` for the log key

docs/tables/github_organization_credential_authorization.md
cbruno10: "`admin:org` scope (specifically `read:org`)" is contradictory. GitHub's docs require `admin:org` for this endpoint. Please also state that it is classic-PAT only, since fine-grained tokens do not support it, and add the permissions block the other org tables carry (see `github_organization_external_identity.md`).

docs/tables/github_organization_credential_authorization.md
```diff
@@ -0,0 +1,209 @@
...
+
+**Important Notes**
+- You must specify the `organization` column in the `where` or `join` clause to query the table.
+- This table is only available for organizations on **GitHub Enterprise Cloud** that have **SAML single sign-on (SSO)** enabled. Organizations without SAML SSO return no rows.
```
cbruno10, line 19: This is only true because of the `403` in `ShouldIgnoreError`. Once that's out, this should say the query errors if the token isn't an org owner or SAML isn't enabled.

github/plugin.go
```diff
@@ -23,74 +23,75 @@ func Plugin(ctx context.Context) *plugin.Plugin {
...
+			"github_issue_comment":                         tableGitHubIssueComment(),
+			"github_license":                               tableGitHubLicense(),
+			"github_my_gist":                               tableGitHubMyGist(),
+			"github_my_issue":                              tableGitHubMyIssue(),
+			"github_my_organization":                       tableGitHubMyOrganization(),
+			"github_my_repository":                         tableGitHubMyRepository(),
+			"github_my_star":                               tableGitHubMyStar(),
+			"github_my_team":                               tableGitHubMyTeam(),
+			"github_organization":                          tableGitHubOrganization(),
+			"github_organization_credential_authorization": tableGitHubOrganizationCredentialAuthorization(),
```
cbruno10, line 52: This is where the merge conflict comes from. The new key is the longest in the map, so `gofmt` shifted every line and anything added on `main` since collides. Can you please take `main`'s `plugin.go`, add just this one line, and then run `gofmt`.

cbruno10 (changes requested):
@Kanthi-Hegde I've left a few suggestions, can you please have a look? Thanks!

PR #556, opened 2026-07-14
Add Custom Properties for Repository-Derived Tables

Recurzion:
This PR adds the ability to query repository custom properties (returned as `jsonb`) for any repository-derived columns via `sharedRepositoryColumns()` and updates docs with example queries (these can be removed if not needed).

Teams often use GitHub custom properties to capture custom repo-level metadata/attributes, so adding this field should benefit organizations that leverage it.

This applies to the following tables:

- github_repository
- github_my_repository
- github_team_repository
- github_search_repository
- and any other table using `sharedRepositoryColumns()`

# Example query results
<details>
  <summary>Results</summary>
  
```

select name,custom_properties  from github_my_repository where name = 'myrepo'

+----------+-------------------------------+
| name | custom_properties                 |
+----------+-------------------------------+
|myrepo |  {"key1":"value1","key2":"value2"}
+----------+-------------------------------+

select name,custom_properties -> 'myproperty' as myproperty from github_repository where  full_name = 'myorg/myrepo'

+----------+-------+
| name | myproperty|
+----------+-------+
|myrepo | "myvalue"|
+----------+-------+



```
</details>

### DIFF
--- a/docs/tables/github_search_repository.md
+++ b/docs/tables/github_search_repository.md
@@ -136,6 +136,31 @@ where
   query = 'tinyspotifyr in:name created:2021-01-01..2021-01-05 fork:only';
 ```
 
+### Get custom property values from search results
+Explore custom property key/value pairs for repositories returned by a search query.
+
+```sql+postgres
+select
+  name_with_owner,
+  custom_properties
+from
+  github_search_repository
+where
+  query = 'org:my-org archived:false'
+  and custom_properties is not null;
+```
+
+```sql+sqlite
+select
+  name_with_owner,
+  custom_properties
+from
+  github_search_repository
+where
+  query = 'org:my-org archived:false'
+  and custom_properties is not null;
+```
+
 ```sql+sqlite
 select
   name,
--- a/docs/tables/github_search_repository.md
+++ b/docs/tables/github_search_repository.md
@@ -136,6 +136,21 @@ where
   query = 'tinyspotifyr in:name created:2021-01-01..2021-01-05 fork:only';
 ```
 
+```sql+sqlite
+select
+  name,
+  owner_login,
+  primary_language,
+  fork_count,
+  stargazer_count,
+  subscribers_count,
+  watchers_total_count
+from
+  github_search_repository
+where
+  query = 'tinyspotifyr in:name created:2021-01-01..2021-01-05 fork:only';
+```
+
 ### Get custom property values from search results
 Explore custom property key/value pairs for repositories returned by a search query.
 
@@ -159,19 +174,4 @@ from
 where
   query = 'org:my-org archived:false'
   and custom_properties is not null;
-```
-
-```sql+sqlite
-select
-  name,
-  owner_login,
-  primary_language,
-  fork_count,
-  stargazer_count,
-  subscribers_count,
-  watchers_total_count
-from
-  github_search_repository
-where
-  query = 'tinyspotifyr in:name created:2021-01-01..2021-01-05 fork:only';
 ```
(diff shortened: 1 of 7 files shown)


github/table_github_repository.go
```diff
@@ -146,6 +157,47 @@ func tableGitHubRepositoryList(ctx context.Context, d *plugin.QueryData, h *plug
 	return nil, nil
 }
 
+type repositoryCustomPropertyValue struct {
+	PropertyName string      `json:"property_name"`
+	Value        interface{} `json:"value"`
+}
+
+func hydrateRepositoryCustomPropertiesFromV3(ctx context.Context, d *plugin.QueryData, h *plugin.HydrateData) (interface{}, error) {
```
cbruno10: Instead of using the v3 RESTful API, can we retrieve it using the newer GraphQL API, similar to most other columns in this table? GraphQL APIs in general are more forgiving with rate limits and from what we could see, better supported moving forward.

In the GraphQL API, the `Repository` object has `repositoryCustomPropertyValues`, so we can use that I believe and we should get the `first: 100`, as there seems to be a limit of 100 property definitions.

You can look at the `labels` column in the `github_issue` table, as it has the same shape and also does an array to map conversion like this property would use.

github/table_github_repository.go
```diff
@@ -19,6 +19,16 @@ func gitHubRepositoryColumns() []*plugin.Column {
 	return append(repoColumns, sharedRepositoryColumns()...)
 }
 
+func repositoryCustomPropertiesColumn() *plugin.Column {
```
cbruno10: Do we need this helper function, especially since it's only for 1 column?

Though if we convert the column data to come from GraphQL instead, I think this helper function would be unnecessary anyway.

docs/tables/github_search_repository.md
```diff
@@ -136,6 +136,31 @@ where
   query = 'tinyspotifyr in:name created:2021-01-01..2021-01-05 fork:only';
 ```
 
+### Get custom property values from search results
```
cbruno10, line 154: I think this example was inserted between an existing postgres and sqlite example, can you please move it so it doesn't breakup the `### List forked repositories created within specific timestamp` example?

docs/tables/github_team_repository.md
```diff
@@ -65,7 +65,6 @@ where
   organization = 'my_org'
   and slug = 'my-team';
 ```
-
```
cbruno10, line 68: Any reason this was removed?

docs/tables/github_team_repository.md
```diff
@@ -105,4 +104,32 @@ where
...
+  slug,
+  name,
+  custom_properties
+from
+  github_team_repository
+where
+  organization = 'my_org'
+  and slug = 'my-team';
+```
+```sql+sqlite
```
cbruno10, line 126:
```suggestion
```

```sql+sqlite
```

cbruno10 (changes requested):
@Recurzion Thanks for the PR!

I've added a couple suggestions, the bigger one is around moving from v3 (REST) to v4 (GraphQL) API, happy to discuss more if you have any questions. Thanks!

docs/tables/github_search_repository.md
Recurzion: Yep, done!

docs/tables/github_team_repository.md
Recurzion: Fixed the spacing here.

github/table_github_repository.go
Recurzion: Sure thing. I followed the same pattern used by the labels column now.

github/table_github_repository.go
Recurzion: Helper function removed now that we're using the GraphQL endpoint instead.

cbruno10 (approved):

